PR #1, opened 2023-02-14
Initialize luxonis-ml

conorsim:
Adds the following:
- `LuxonisTracker`
- `LuxonisDataset`
- `LuxonisDatasetArtifact`
- `LuxonisLoader`
- `luxonis_ml` CLI tool


README.md
```diff
@@ -0,0 +1,424 @@
+## Luxonis Dataset Format
```
tersekmatija, line 1: Do you mind just adding Alpha Version notice at the top. Something saying like that the API might change.

src/luxonis_ml/ops/artifacts.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def pull(self, s3_pull=False):
+
+        if hasattr(self.dataset, 'commit_id'):
+            t0 = time.time()
+
+            if s3_pull:
+                print("Running rclone to sync with S3...")
+                bucket_path = self.dataset.s3_path.split('s3://')[-1]
+                cmd = f"rclone sync --s3-force-path-style=true lakefs:{self.name}/{self.dataset.commit_id}/ b2:{bucket_path}/{self.dataset.bough.value}"
```
tersekmatija, line 236: Would it make sense to keep b2 as a variable somewhere in case we change the provider?

src/luxonis_ml/ops/dataset.py
```diff
@@ -0,0 +1,740 @@
...
+
+    def _get_credentials(self, key):
+        if key in self.creds.keys():
+            return self.creds[key]
+        else:
+            return os.environ[key]
+
+    def _init_boto3_client(self):
+        self.client = boto3.client('s3',
+                        endpoint_url = 'https://s3.us-west-004.backblazeb2.com',
```
tersekmatija, line 214: Let's have this as a variable too.

src/luxonis_ml/ops/artifacts.py
```diff
@@ -0,0 +1,267 @@
...
+
+                    local_dir = relative_path.split(relative_path.split('/')[-1])[0]
+                    if not local_dir.startswith(self.dataset.bough.value):
+                        local_dir = f"{self.dataset.bough.value}/{local_dir}"
+                    os.makedirs(f"tmp/{local_dir}", exist_ok=True)
+                    cmd = f"cp {relative_path} tmp/{local_dir}"
+                    subprocess.check_output(cmd, shell=True)
+
+                print("Uploading data to S3 - this could take some time...")
+                cmd = f"b2 sync tmp/{self.dataset.bough.value} b2://{self.dataset.bucket}/lakefs/{self.name}/_ldf_commits/{ldf_commit_number}/"
```
tersekmatija, line 186: Let's keep it like that for now. Might be worth testing with some other more general CLI that can allow us to provide custom endpoints. In case we switch providers.

tersekmatija: Would be good to mention this dependency on b2 somewhere in the README.

src/luxonis_ml/ops/dataset.py
```diff
@@ -0,0 +1,740 @@
...
+
+                self.add_data(source_name, {
+                    component_name: img,
+                    'json': new_ann
+                }, split=split)
+
+            else:
+                warnings.warn(f"skipping {fn} as it does no exist!")
+
+    def from_yolo(self, source_name, yaml_path, split='all', override_main_component=None):
```
tersekmatija, line 660: As we grow, might be good to move those to separate files. Something like "parsers". To keep the readability of dataset and to allow easy support of additional parsers.

tersekmatija (changes requested):

src/luxonis_ml/ops/artifacts.py
conorsim: This is now a variable in the `pull` function. Good call, as it depends on `rclone config`. It is also now an option in `luxonis_ml` CLI

src/luxonis_ml/ops/dataset.py
conorsim: Fixed

src/luxonis_ml/ops/artifacts.py
conorsim: We can also try using rclone for this. I think I found that b2 was faster, but I could explore this more. For now, I'll keep it this way and mention it in the README.

src/luxonis_ml/ops/dataset.py
conorsim: Split these into a new file

src/luxonis_ml/ops/artifacts.py
tersekmatija: Sounds good. Do you remember what the comparison between b2 and rclone was? @conorsim

README.md
```diff
@@ -0,0 +1,461 @@
...
+
+#### Adding Data
+
+For COCO and YOLO, it's quite easy to add data with the `from_coco` and `from_yolo` methods.
+
+```python
+with LuxonisDataset("repo") as dataset:
+
+    # add COCO 2017 val split
+    dataset.from_coco(
```
tersekmatija, line 230: I believe this should be updated now? @conorsim

tersekmatija (approved):
Only the docs should be update now I think. Otherwise looks good. And once we have some tests with b2 vs rclone, we can do an update.

src/luxonis_ml/ops/artifacts.py
conorsim: On uploading COCO val:
`b2`: about 11 minutes
`rclone`: about 24 minutes

PR #2, opened 2023-03-01
Fix bug which prevents local use

tersekmatija:
Replaces `and` with `or` in the following line:
```
if s3_path is None and local_path is None:
```


Erol444 (approved):
lgtm

PR #5, opened 2023-03-03
loader fix for datasets w/o keypoint annotations

jkbmrz:
fix the loader.py to also run for datasets without keypoint annotations


tersekmatija (approved):
LGTM!

PR #6, opened 2023-03-23
Adding custom from_xyz functions used to load data of different formats.

jkbmrz:


tersekmatija:
I've reviewed offline, do you mind quickly throwing an eye on it too when you get a sec @conorsim ?

PR #7, opened 2023-04-02
To webdataset fix

jkbmrz:
I noticed that to_webdataset() method has exponential runtime depending on the amount of data instances in our LDF dataset. This problem becomes very obvious when running on Google Colab which has fairly slow file/folder management.

Thus, I propose this change to the code which eliminates usage of glob.glob() search for every basename in our dataset. Instead, it extracts file extensions for a single basename and assumes that other basenames come with same extensions. I believe this is a meaningful assumption, based on how we define data sources and components.

We definitely need to fix this exponential runtime problem but maybe there exists a more elegant way that what I propose (that's why I propose it through a PR).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,5 @@ pandasql>=0.7.3
 pycocotools>=2.0.6
 pytorch-lightning>=1.9.0
 boto3>=1.17.39
+sqlalchemy==1.4.46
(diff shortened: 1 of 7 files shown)


jkbmrz:
Whops, made a mistake. I want to merge with 'dev' branch.

PR #12, opened 2023-07-10
S3 boto3, filename hashing, update fixes

tersekmatija:
Replace awscli with boto3 and multithreading, added filename hashing based on file content, version stacked UPDATE transactions when multiple transactions change the same sample.

### DIFF
--- a/src/luxonis_ml/data/dataset.py
+++ b/src/luxonis_ml/data/dataset.py
@@ -20,6 +20,7 @@ from datetime import datetime
 import pymongo
 from bson.objectid import ObjectId
 from .version import LuxonisVersion
+from .utils.s3_utils import sync_from_s3, sync_to_s3
 
 class HType(Enum):
     """ Individual file type """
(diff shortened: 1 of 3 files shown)


src/luxonis_ml/data/dataset.py
```diff
@@ -638,7 +637,7 @@ def _add_extract(self, additions, from_bucket):
                         Key=new_prefix,
                         CopySource={'Bucket': self.bucket, 'Key': old_prefix}
                     )
-                else:
+                elif filepath != "{local_cache}/{component_name}/{granule}":
```
conorsim, line 640: @tersekmatija I wonder if we can use `media_exists` here instead or find the cause of the `cp` errors since this seems like a workaround.

Actually, maybe we should even remove the `media_exists` entirely and only upload/extract the data in the case that the unique hash file does not exist. We can use `os` or `boto3` for this.

src/luxonis_ml/data/dataset.py
```diff
@@ -701,7 +696,13 @@ def _add_execute(self, additions=None, transaction_to_additions=None):
                 if additions is None or transaction_to_additions is None:
                     raise Exception("additions and transaction_to_additions required for adding data")
 
-                addition = additions[transaction_to_additions[transaction['_id']]]
+                if transaction_to_additions.get(str(transaction['_id'])) is None:
+                    # TODO: find a better way to handle those
+                    print(f"Cannot find transaction with id: {str(transaction['_id'])}")
+                    print(f"Seems like a failed transaction?")
+                    continue
```
conorsim, line 703: Let's wait to see if this is still an issue after adding some rollback features because I also encountered this when I would get errors in `_add_filter` etc. and there isn't proper error handling of the transactions

conorsim (approved):
Comments were more for discussion, LGTM at the moment

PR #15, opened 2023-07-11
Small features and fixes

conorsim:
- Ensure all components have the same split
- Add an `instance_id` field to all samples with the instance hash
- Check format for bounding box label to start with class
- Replace static paths with pathlib for paths referencing local OS
- If no field is present in an update, assume it does not change instead of assuming it is removed. This also fixes the issue with `split` becoming `None`
- Add `ldf_version` to the LDF mongo doc
- Rework bucket_type with `BucketType` and `BucketStorage`

### DIFF
--- a/src/luxonis_ml/data/dataset.py
+++ b/src/luxonis_ml/data/dataset.py
@@ -148,8 +148,8 @@ class LuxonisDataset:
         if self.bucket_type not in self.bucket_choices:
             raise Exception(f"Bucket type {self.bucket_type} is not supported!")
 
-        credentials_cache_file = (
-            f"{str(Path.home())}/.cache/luxonis_ml/credentials.json"
+        credentials_cache_file = str(
+            Path.home() / ".cache" / "luxonis_ml" / "credentials.json"
         )
         if os.path.exists(credentials_cache_file):
             with open(credentials_cache_file) as file:
@@ -272,7 +272,15 @@ class LuxonisDataset:
 
     def _init_path(self):
         if self.bucket_type == "local":
-            self.path = f"{str(Path.home())}/.cache/luxonis_ml/data/{self.team_id}/datasets/{self.dataset_id}"
+            self.path = str(
+                Path.home()
+                / ".cache"
+                / "luxonis_ml"
+                / "data"
+                / self.team_id
+                / "datasets"
+                / self.dataset_id
+            )
             os.makedirs(self.path, exist_ok=True)
         elif self.bucket_type == "aws":
             self.path = f"s3://{self._get_credentials('AWS_BUCKET')}/{self.team_id}/datasets/{self.dataset_id}"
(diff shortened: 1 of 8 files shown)


src/luxonis_ml/data/dataset.py
```diff
@@ -125,16 +146,9 @@ def __init__(
         dataset_id,
         team_name=None,
         dataset_name=None,
-        bucket_type="local",
-        override_bucket_type=False,
+        bucket_type=BucketType.EXTERNAL,
```
tersekmatija, line 149: How do external + local work together?

src/luxonis_ml/data/utils/data_utils.py
```diff
@@ -102,11 +48,22 @@ def check_classification(val1, val2):
...
-    if val2 is None and val1 is not None:
+    if (val1 is None and val2 is not None) or (val2 is None and val1 is not None):
         return [{"boxes": val1}]
+    elif val1 is None and val2 is None:
+        return []
 
     if len(val1) == len(val2["detections"]):
         for val1, val2 in list(zip(val1, val2["detections"])):
+            # assert bounding boxes contain the right format of either int and list or str and list
+            if not (
```
tersekmatija, line 59: Should we maybe also add x1,y1,w,h check? I assumed it should be in Yolo format and then training was failing. I bet I won't be the only one so perhaps validating this here would still be worth it.

src/luxonis_ml/data/dataset.py
```diff
@@ -631,13 +661,26 @@ def _add_filter(self, additions):
...
-                            sample_id=latest_sample._id,
-                            field=field,
-                            value=value,
-                            component=component_name,
-                        )
+                        if field == "split":
+                            group = data_utils.get_group_from_sample(
+                                self, latest_sample
+                            )
+                            for component_name in group:
```
tersekmatija, line 668: Does this make the updates slower?

src/luxonis_ml/data/utils/data_utils.py
```diff
@@ -275,11 +245,24 @@ def generate_hashname(filepath):
...
         return False
+
+
+def get_group_from_sample(dataset, sample):
+    group = sample[dataset.source.name]
+    group = dataset.fo_dataset.get_group(group["id"])
+    return group
+
+
+def get_filepath_from_hash(dataset, hash):
```
tersekmatija, line 264: What if instance wouldn't exist? Would it make sense to output some generic path in that case?

tersekmatija (changes requested):
LGTM, requested changes to pay attention to the comments before merging.

src/luxonis_ml/data/utils/data_utils.py
conorsim: Can we actually distinguish between xywh and xyxy? Is that the check you mean? 

Or do you mean checking for float and 0-1 range or something like that?

src/luxonis_ml/data/dataset.py
conorsim: Yes it will. I think probably not by much though and only in the case where split is updated.

src/luxonis_ml/data/utils/data_utils.py
conorsim: I just edited it to return `None` now in that case

tersekmatija: Should we throw some error if it happens?

src/luxonis_ml/data/utils/data_utils.py
tersekmatija: > Or do you mean checking for float and 0-1 range or something like that?

Yeah, and for x1,y1,w,h I meant just some basic checks. Like we know x1+w < 1, where as if x1 and x2 are passed it could happen it's higher. Like we can't check all, but if there is such scenario then it's preventable basically.

src/luxonis_ml/data/dataset.py
tersekmatija: Yeah I agree, if local it shouldn't really matter if internal or external. I'd still default to internal. We'll use the external flag to mark that we are not the owners. Which will be for BYOB.

PR #16, opened 2023-07-13
Initial Exceptions

conorsim:
- Efficiently store segmentation arrays as serialized bytes (using fiftyone `ArrayField`) to speed up transactions
- Custom exception handling for `LuxonisDataset.add`, which provide more detailed error messages
- Rollback for `LuxonisDataset.add` so that in the case of an error or `KeyboardInterrupt`, we attempt to return the dataset to its original state
- Thorough annotation checking for classification, bounding boxes, segmentation, and keypoints to ensure the correct format from `additions`

### DIFF
--- a/src/luxonis_ml/data/utils/data_utils.py
+++ b/src/luxonis_ml/data/utils/data_utils.py
@@ -18,36 +18,75 @@ def get_granule(filepath, addition, component_name):
     return granule
 
 
+def assert_classification_format(val):
+    if val is not None:
+        if isinstance(val, str):
+            pass
+        elif isinstance(val, list):
+            for v in val:
+                if not isinstance(v, str):
+                    raise ClassificationFormatError(
+                        "All elements in list must be string"
+                    )
+        else:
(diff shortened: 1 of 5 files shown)


src/luxonis_ml/data/utils/data_utils.py
```diff
@@ -18,52 +18,131 @@ def get_granule(filepath, addition, component_name):
     return granule
 
 
-def check_classification(val1, val2):
+def assert_classification_format(dataset, val):
+    if val is not None:
+        if isinstance(val, str):
+            if val not in dataset.fo_dataset.classes.get("class", []):
+                raise ClassNotFoundError(f"Class {val} is not found in dataset")
```
tersekmatija, line 25: Maybe unknown class?

tersekmatija: "ClassUnknownError" for example. Even better might be if we stick with "Exception"?

tersekmatija (changes requested):

PR #18, opened 2023-07-26
Augmentations refactor

klemen1999:
Refactor of Augmentations class in order to support batched transforms (e.g. Mosaic)
- Refactored Augmentations to have `is_batched` and `batch_size` attributes (automatically defined based on type of augmentations defined in input dict. If augmentation `is_batched` then pass N images from LuxonisLoader to Augmentations (N=`batch_size`)
- User defined augmentations are split into single and multi image ones and based on that we define `A.BatchCompose()` - single image augs go in `A.ForEach()`, multi image augs go in seperate step (if present)
- After user defined augmentations are done we always do a separate augmentation to resize previous output

**Note:** BatchCompose is not yet part of official Albumentations so we changed path in requirements to point to specific fork. 

Other changes:
- Introduced `LabelType` for better code managment, we use this also in `models`
- Small change in tracker.py where we set `project_id` and `run_id` of the current run if using MLFlow (used in `models`)


tersekmatija:
Thanks, @jkbmrz @conorsim please review.

Quick questions from my side @klemen1999 after just briefly reading through:
* The batch_size only refers to the batch processed by augmentations, right? So in case of mosaic, we can still have a larger training batch size.
* In case of mosaic, does each image get separately augmented?
* Do we know whether the fork will be merged? Is there a PR waiting for the review already? Otherwise if not mainlined there's a risk it might get stale and outdated.

klemen1999:
1/ Yes this only refers to the batch size needed for augmentation and it doesn't affect the training batch size. It takes N-1 other random samples (with replacements if the dataset is too small). Maybe it would be better to rename it to something like `aug_batch_size` to avoid confusion. 
2/ First all 4 images go through augmentations that are for single image (e.g. Rotate, Blur, Crop,...) and then in last step the Mosaic is performed.
3/ There is a [PR](https://github.com/albumentations-team/albumentations/pull/1420) for this fork and there was a recent merge of master into this branch by albumentations team member so it looks like they are planning to add it to official release.

src/luxonis_ml/loader/augmentations.py
```diff
@@ -33,82 +49,219 @@ def _parse_cfg(
...
-
-        return A.Compose(
-            all_augs,
-            bbox_params=A.BboxParams(format="coco", label_fields=["bbox_classes"]),
+                if aug["name"] in ["Mosaic4"]:
+                    self.is_batched = True
+                    self.batch_size = 4
+                    multi_image_augs.append(eval(aug["name"])(**aug.get("params", {})))
+                else:
+                    single_image_augs.append(eval(aug["name"])(**aug.get("params", {})))
```
conorsim, line 71: Assuming we will have multiple multi-image augmentations, should we make some enum for single image augs, multi-image augs, or both?

src/luxonis_ml/loader/loader.py
```diff
@@ -1,18 +1,30 @@
...
 import numpy as np
 import torch
+import random
+import warnings
+from pathlib import Path
 from fiftyone import ViewField as F
+from enum import Enum
+
+
+class LabelType(str, Enum):
```
conorsim, line 11: I like `LabelType`. I think we could actually define it in the `data` module and use it there as well. I think it will be a new feature but if we agree I'll note it down.

conorsim (approved):

jkbmrz:
The code looks good to me!

I am thinking though, if applying mosaic augmentation in the beginning or the end of augmentation process has any dis/advantages. For example, when applying translations (moving images left/right/up/down) before mosaic augmentation, the output images constantly lack information in the middle of the image, the place where the model should be the most focused.

jkbmrz (approved):

src/luxonis_ml/loader/augmentations.py
```diff
@@ -33,82 +49,219 @@ def _parse_cfg(
...
+            keypoints_batch=keypoints_batch,
+            keypoints_visibility_batch=keypoints_visibility_batch,
+            keypoints_classes_batch=keypoints_classes_batch
+        )
+
+        # convert to numpy arrays
+        for key in transformed:
+            transformed[key] = np.array(transformed[key][0])
+
+        transformed = self.default_transform(
```
conorsim, line 135: I meant to comment on this earlier but since `default_transform` seems to happen last, but we automatically adjust the resize to be larger in the case of `Mosaic4`? For example, we'd want width and height x2 compared to the normal training size. And the user might not remember to adjust this manually in the training config.

tersekmatija:
> For example, when applying translations (moving images left/right/up/down) before mosaic augmentation, the output images will constantly lack information in the middle of the image, the place where the model should be the most focused.

I think translations for specific image should not be applied in that scenario. Do we still apply them?

src/luxonis_ml/loader/loader.py
klemen1999: Yeah I was thinking about using it in `data` module as well. But I wasn't sure how to then still keep modules separate because if we define it in `data` then we would need to import `data` in `loader`. Should there maybe be a separate file with constants outside of modules with shared types?

src/luxonis_ml/loader/augmentations.py
klemen1999: Not sure what you mean, can you axplain a bit more? I know that this `if name in []` is not scalable solution but I didn't have an idea of how to do it otherwise. Maybe we first evaluate the augmentation and then check if is of type BatchBasedTransform (based on the [source code](https://github.com/i-aki-y/albumentations/blob/c137e2b8bb48eb25b8865db80027f16059724db7/albumentations/augmentations/geometric/transforms.py#L1502)). And in this case we could also check [mosaic.n_tiles ](https://github.com/i-aki-y/albumentations/blob/c137e2b8bb48eb25b8865db80027f16059724db7/albumentations/augmentations/geometric/transforms.py#L1548C9-L1548C21) to define self.batch_size although this is the only implemented BatchedTransform so far so not sure if others will also follow this class signature.

src/luxonis_ml/loader/augmentations.py
klemen1999: There are `out_height `and `out_width ` parameters for `Mosaic4 `that have to be defined by the user ([this](https://github.com/i-aki-y/albumentations/blob/c137e2b8bb48eb25b8865db80027f16059724db7/albumentations/augmentations/geometric/transforms.py#L1510-L1515)). So if we would want bigger crops of images in each quadrant this two values should be high (e.g. for full image in each quadrant this should be 2xH by 2xW). But this only works if all images have same size (not true in COCO for example). 

Do we want to automatically x2 shape if this parameters are not set by the user? And if so, where should we handle this: in Augmentations or before, when loading config in `models` for example?

klemen1999:
In current implementation every single image augmentation defined by the user happens before Mosaic. So in your example @jkbmrz if you specified translation augmentation it would be applied to each image and then all the outputs would be merged to Mosaic. Is this the desired outcome?

If we would want to translate the output of Mosaic we would need to do it in separate step (as we do now with final resize) but then we would need to somehow add support in config for augmentations that we want to perform before and after BatchTransform so we can sort them accordingly when parsing.

tersekmatija:
> If we would want to translate the output of Mosaic we would need to do it in separate step (as we do now with final resize) but then we would need to somehow add support in config for augmentations that we want to perform before and after BatchTransform so we can sort them accordingly when parsing.

@klemen1999 My guess is we could be more limiting here. I think we could define which operations are executed before and which after batch transformations for now. And if batch transformations are enabled and this augmentation is present, we execute it based on those pre-defined rules. For example, everything that modifies image "pixels" could be before. Translations and warping for example could be after. I think the main issue with translation before mosaic is that end images would have a lot of empty space between them. Whereas translation after mosaic just means we shift the image slightly away from the center.

jkbmrz:
Indeed, applying pixel-level augmentations first will allow for more variation and diversity within each of the four images that make up the mosaic (hopefully improving the training). Spatial-level augmentations could be done after (it might be more realistic as we don't expect flipping/rotating/transposing only on parts of the image in real world scenarios).

klemen1999:
Agree. Then I think we can sort them based on type and we do it in this order: 
- First we do augmentations of type ImageOnlyTransform (includes all pixel level augmentations)
- Then we do BatchBasedTransform
- Then DualTransform (includes all spatial level transforms)
- And lastly Resize to resize into training shape

src/luxonis_ml/loader/augmentations.py
conorsim: Yeah I think this solution is fine for now and we could wait to see if we use any other batched transforms first. I originally meant we could define an enum that would match to the augmentation names like

```python
class MultiImageAugmentation(Enum):
    MOSAIC = "Mosaic4"
    ...
```
and then a check like
```python
if aug["name"] in MultiImageAugmentation:
```

If we can check for `BatchBasedTransform` it might be even cleaner

src/luxonis_ml/loader/loader.py
conorsim: Oh yeah maybe we can move all the enums to some enums folder? Or something like constants. I saw @tersekmatija did this in some project.

klemen1999: In `models` I have them under utils/constants.py but that is a single module. So yeah here we could do something similar, just outside of specific module.

tersekmatija:
CC @jkbmrz @conorsim

src/luxonis_ml/loader/loader.py
```diff
@@ -100,7 +100,7 @@ def load_image_with_annotations(self, idx: int):
         if self.stream and self.dataset.bucket_storage.value != "local":
             img_path = str(Path.home() / ".luxonis_mount" / path[1:])
         else:
-            img_path = f"{str(Path.home())}/.cache/luxonis_ml/data/{path}"
+            img_path = str(Path.home() / ".cache" / "luxonis_ml" / "data" / path[1:])
```
jkbmrz, line 103: Wouldn't it be safer to do the path construction with os.path.join?

conorsim: I think this is also safe. It would be nice to stick to one or the other though maybe but not sure.

conorsim (approved):
Looks good

PR #22, opened 2023-08-29
GCS Support, annotated tag, delete rework

conorsim:
- Add support for GCS in both `from_bucket` and from laptop
- Add option when calling `dataset.add()` to specify `annotated=False` to ignore this in versioning for the FE until it is annotated
- Rework updating attributes based on instance_id and add a function to annotate an instance once annotated

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -608,7 +608,7 @@ class LuxonisDatasetTester(unittest.TestCase):
         with LuxonisDataset(self.team_id, self.dataset_id) as dataset:
             for sample in dataset.fo_dataset:
                 break
-            dataset.delete(["fe163bd9-6381-5533-9d41-c1735edf96d5"])
+            dataset.delete([sample.instance_id])
 
             res = (
                 dataset._check_transactions_to_version()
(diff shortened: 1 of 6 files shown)


tersekmatija:
Please review in the morning @kozlov721 . I'll try to review this tomorrow as well @conorsim .

tests/test.py
```diff
@@ -608,10 +608,10 @@ def test_delete(self):
         with LuxonisDataset(self.team_id, self.dataset_id) as dataset:
             for sample in dataset.fo_dataset:
                 break
-            dataset.delete([sample.id])
+            dataset.delete(["fe163bd9-6381-5533-9d41-c1735edf96d5"])
```
kozlov721, line 611: Is this supposed to be hardcoded in?

conorsim: It was intentional since it should always be the same based on image bytes. However, I could also find this like I did with sample ID to ensure there are no issues with it being hardcoded.

conorsim: @kozlov721 latest commit removes this hardcode

kozlov721 (approved):
lgtm

PR #23, opened 2023-09-01
Embeddings based methods

DrejcPesjak:
Scripts for generating embeddings from images in LDF using pytorch or ONNX models, inserting the embeddings into a Qdrant database (vector similarity search engine) and executing embedding based methods, such as near-duplicate search, mismatch detection in labeled data, out-of-distribution (OOD) detection, and identifying representative images.

The full usage examples are in _examples/test_emb_process.ipynb_ and _examples/test_ldf_emb.ipynb_.

Maybe one question is if I should put the functions into classes, especially qdrant_utils.py could be split at line 112 into QdrantManager (qdrant docker initialization) and QdrantAPI (qdrant querying - because all these functions have qdrant_client and collection_name as a parameter).


src/luxonis_ml/embeddings/emb_OOD.py
```diff
@@ -0,0 +1,99 @@
...
+    X : np.array
+        The embeddings to use.
+    contamination : float
+        The contamination parameter for Isolation Forests. Default is 'auto'.
+    n_jobs : int
+        The number of jobs to use. Default is -1, which means all available CPUs.
+    verbose : int
+        The verbosity level. Default is 1.
+    random_state : int
+        The random state to use. Default is 42.
```
conorsim, line 48: Should the default `random_state` be `None`? I'm not sure. I could see it being nice if the default behavior is such that the result is reproducible.

src/luxonis_ml/embeddings/emb_OOD.py
```diff
@@ -0,0 +1,99 @@
...
+        The indices of the embeddings that are out-of-distribution.
+
+    """
+    # Calculate the hat matrix (projection matrix) to get the leverage for each point
+    hat_matrix = np.matmul(np.matmul(X, np.linalg.inv(np.matmul(X.T, X))), X.T)
+    leverage = np.diagonal(hat_matrix)
+
+    # Calculate the leverage threshold using 3 standard deviations
+    mean, std = np.mean(leverage), np.std(leverage)
+    upper_threshold, lower_threshold = mean + 3 * std, mean - 3 * std
```
conorsim, line 95: Should the 3 standard deviations here be a configurable parameter?

src/luxonis_ml/embeddings/emb_duplicate.py
```diff
@@ -0,0 +1,273 @@
...
+# Qdrant
+from qdrant_client import QdrantClient
+from qdrant_client.http import models
+
+from luxonis_ml.embeddings.qdrant_utils import *
+
+
+# client = QdrantClient(host="localhost", port=6333)
+
+def search_qdrant(qdrant_client, query_vector, collection_name="webscraped_real_all", data_name="flash", limit=5000):
```
conorsim, line 80: It seems like the `collection_name` and `data_name` should be required args to me

conorsim: (Same for other functions that have this)

src/luxonis_ml/embeddings/emb_duplicate.py
```diff
@@ -0,0 +1,273 @@
...
+                    minima = max(minima, 0.96)
+                
+            else:
+                minima = 0.97
+
+            # select k closest embeddings
+            k = len(np.where(similarities > minima)[0])
+            if minima > 0.97 and k < 2:
+                k = len(np.where(similarities > 0.97)[0])
+            print(k, minima)
```
conorsim, line 265: There are a lot of constants here. I think it's fine to keep them as is for now if it seems to work well, but we might want to experiment with changing them and if they should be changed by a user in the future.

src/luxonis_ml/embeddings/emb_mistakes.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Find mismatches in the dataset. 
+    A mismatch is defined as a sample that is closer to another centroid than to its own centroid.
+
+    Parameters
+    ----------
+    X : np.array
+        The embeddings to use.
+    y : np.array
+        The targets to use.
```
conorsim, line 48: More of a question, but is `y` assumed to always be a single label per image? Does this method only apply to labeled data for image classification?

src/luxonis_ml/embeddings/emb_mistakes.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Get the embeddings of the samples in the specified class
+        embeddings_class = X[indices]
+
+        # Calculate the centroid of the specified class
+        centroid = np.mean(embeddings_class, axis=0)
+        centroids.append(centroid)
+
+    # append the centroids to the embeddings
+    Centroids = np.array(centroids)
```
conorsim, line 75: This variable being capitalized makes me think it's a special class, but not a big deal

src/luxonis_ml/embeddings/emb_representative.py
```diff
@@ -0,0 +1,170 @@
...
+"""
+
+from sklearn.metrics.pairwise import cosine_similarity
+from kmedoids import KMedoids
+
+from luxonis_ml.embeddings.qdrant_utils import *
+
+
+def calculate_similarity_matrix(embeddings):
+    return cosine_similarity(embeddings)
```
conorsim, line 51: Do we need this wrapper?

src/luxonis_ml/embeddings/ldf_emb.py
```diff
@@ -0,0 +1,292 @@
...
+import torchvision
+import torchvision.transforms as transforms
+from qdrant_client import QdrantClient
+from qdrant_client.models import Distance, VectorParams, PointStruct, SearchRequest
+from qdrant_client.http import models
+
+DEFAULT_COLLECTION_NAME = "mnist"
+DEFAULT_OUTPUT_LAYER_NAME = "/Flatten_output_0"
+DEFAULT_EMB_BATCH_SIZE = 64
+DEFAULT_QDRANT_BATCH_SIZE = 64
```
conorsim, line 48: We likely want to avoid these global variables. Maybe a class definition makes sense here to make these more easily configurable but still easy to use across functions?

conorsim: We could also use a `self.dataset` for LDF, similar to `LuxonisLoader`

src/luxonis_ml/embeddings/ldf_emb.py
```diff
@@ -0,0 +1,292 @@
...
+        ort_inputs = {ort_session.get_inputs()[0].name: batch_tensor.cpu().numpy()}
+        ort_outputs = ort_session.run([output_layer_name], ort_inputs)
+        
+        # Append the embeddings from the batch to the new_embeddings list
+        batch_embeddings = ort_outputs[0].squeeze()
+        new_embeddings.extend(batch_embeddings.tolist())
+    
+    return new_embeddings
+
+def _batch_upsert(qdrant_client, new_embeddings, new_payloads, qdrant_batch_size = 64, qdrant_collection_name="mnist"):
```
conorsim, line 220: I would make `qdrant_collection_name` required. Or an attribute of the class

src/luxonis_ml/embeddings/ldf_emb.py
```diff
@@ -0,0 +1,292 @@
...
+            print(e)
+            print(f"Failed to upsert batch {i // qdrant_batch_size + 1} to Qdrant.")
+
+def generate_embeddings(luxonis_dataset, 
+                         ort_session, 
+                         qdrant_client, 
+                         qdrant_collection_name=DEFAULT_COLLECTION_NAME, 
+                         output_layer_name=DEFAULT_OUTPUT_LAYER_NAME,
+                         emb_batch_size=DEFAULT_EMB_BATCH_SIZE, 
+                         qdrant_batch_size=DEFAULT_QDRANT_BATCH_SIZE):
```
conorsim, line 254: Oh I see the global variables are only used here. It seems inconsistent with other functions though

src/luxonis_ml/embeddings/qdrant_utils.py
```diff
@@ -0,0 +1,443 @@
+"""
+Qdrant Docker Management and Embedding Operations
+
+This script provides a set of utility functions to manage Qdrant using Docker and perform various operations related to embeddings.
```
conorsim, line 4: Going back to your question, I think classes here would make sense

conorsim (changes requested):
This is looking great and your comments were very helpful. Mainly requesting some changes to the function definitions. I'm really curious to test this on other datasets.

src/luxonis_ml/embeddings/emb_OOD.py
tersekmatija: Yes I would set this to none and leave it to the user if they want it reproducible.

src/luxonis_ml/embeddings/emb_duplicate.py
tersekmatija: I think it would also make sense to move model_utils and qdrant_utils to new /utils subdir, then add constant.py and define like RELEVANT_THRESHOLD_NAME?

src/luxonis_ml/embeddings/embedding_utils.py
```diff
@@ -0,0 +1,100 @@
...
+                       data_loader: torch.utils.data.DataLoader) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Extract embeddings from the given PyTorch model.
+    """
+    embeddings = []
+    labels = []
+
+    with torch.no_grad():
+        for images, batch_labels in data_loader:
+            outputs = model(images)
```
tersekmatija, line 62: Would it make sense to also specify the output layer / module there? (for example, the module name when you print the model architecture). I could see this being beneficial, but I am also fine with just passing a model that is returning the embeddings directly.

src/luxonis_ml/embeddings/embedding_utils.py
```diff
@@ -0,0 +1,100 @@
...
+            embeddings.extend(outputs.squeeze())
+            labels.extend(batch_labels)
+
+    return torch.stack(embeddings), torch.tensor(labels)
+
+def extract_embeddings_onnx(ort_session: ort.InferenceSession, 
+                            data_loader: torch.utils.data.DataLoader, 
+                            output_layer_name: str = "/Flatten_output_0") -> Tuple[torch.Tensor, torch.Tensor]:
```
tersekmatija, line 70: Let's request this and don't set a default. I think can be different for different models.

src/luxonis_ml/embeddings/ldf_emb.py
```diff
@@ -0,0 +1,292 @@
...
+        transform: Optional torchvision transform for preprocessing images.
+    Returns:
+        List of generated embeddings.
+    """
+    # Generate embeddings for the new images using batching
+    new_embeddings = []
+
+    if transform is None:
+        # Define a transformation for resizing and normalizing the images
+        transform = transforms.Compose([
```
tersekmatija, line 194: Does it make sense setting this by default? What is user does want a None transform (just skip it)?

src/luxonis_ml/embeddings/requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+matplotlib>=3.7.2
+numpy>=1.25.2
+onnx>=1.14.0
+# onnxruntime==1.15.1
+onnxruntime-gpu>=1.15.1
+opencv-python>=4.8.0.76
+qdrant-client>=1.4.0
+scikit-learn>=1.3.0
+scipy>=1.11.1
+torch==1.13.1+cu116
```
tersekmatija, line 13: I would omit PyTorch dependencies here and then make it so that if we install this part of the package, we also need the loader part.

tersekmatija (changes requested):
Thanks, looks good from a quick review apart from the discussed comments.

Another comment I have is regarding the naming of the files. I'd remove `emb_`prefix and put this under /methods dir (or potentially better naming). I'd move utils to /utils.

src/luxonis_ml/embeddings/emb_mistakes.py
DrejcPesjak: Yes, currently its implemented like that. Should it support any other cv tasks?

src/luxonis_ml/embeddings/emb_representative.py
DrejcPesjak: There was an idea behind it. Firstly the user can define his own function for calculating the similarity matrix (like maybe based on some other distance) and also so the user can calculate similarity without separately importing sklearn.

src/luxonis_ml/embeddings/ldf_emb.py
DrejcPesjak: I think I programmed it like that because the default model ResNet50 needs an input size of 224x224.
Or in more general case I believe the transform should never be None, it is recommended to atleast resize all the images to the correct size.

src/luxonis_ml/embeddings/emb_mistakes.py
conorsim: Maybe one day but I don't think so right now - I was just curious

conorsim (approved):
Looks good!

kozlov721 (approved):
LGTM

PR #25, opened 2023-09-26
Letterbox resize

klemen1999:
Added custom `LetterboxResize` augmentation that keeps original aspect ration and fills the remaining space with constant value.
Also reformated `embeddings` directory with black formatter.


klemen1999:
Also includes custom `Mosaic4` implementation that is compatible with our processing

src/luxonis_ml/loader/augmentations.py
```diff
@@ -334,4 +389,197 @@ def __init__(
...
+        params = super().update_params(params, **kwargs)
+
+        img_height = params["rows"]
+        img_width = params["cols"]
+
+        ratio = min(self.height / img_height, self.width / img_width)
+        new_height = int(img_height * ratio)
+        new_width = int(img_width * ratio)
+
+        # only supports center alignment
```
tersekmatija, line 450: Would we want to add support for random position in the future?

src/luxonis_ml/loader/augmentations.py
```diff
@@ -334,4 +389,197 @@ def __init__(
...
+            interpolation=cv2.INTER_NEAREST,
+        )
+        img_out = cv2.copyMakeBorder(
+            resized_img,
+            pad_top,
+            pad_bottom,
+            pad_left,
+            pad_right,
+            cv2.BORDER_CONSTANT,
+            self.mask_value,
```
tersekmatija, line 515: Is 0 usually the background class by default?

src/luxonis_ml/loader/augmentations.py
```diff
@@ -334,4 +389,197 @@ def __init__(
...
+        pad_left: int,
+        pad_right: int,
+        **params
+    ) -> BoxInternalType:
+        x_min, y_min, x_max, y_max = denormalize_bbox(
+            bbox, self.height - pad_top - pad_bottom, self.width - pad_left - pad_right
+        )[:4]
+        bbox = [x_min + pad_left, y_min + pad_top, x_max + pad_left, y_max + pad_top]
+        # clip bbox to image, ignoring padding
+        for i in range(4):
```
tersekmatija, line 534: I assume albumentations expects list of floats and not a numpy array?
Otherwise I think np.clip(..., min=[pad_left, pad_top, pad_left, pad_top], max=...) should be faster.

tersekmatija (approved):
Thanks, some small changes requested otherwise looks good to me.

Please keep the formatter PRs totally separate from the rest of PRs. Makes it really hard to review ;)

src/luxonis_ml/loader/augmentations.py
klemen1999: Albumentations [PadIfNeeded](https://albumentations.ai/docs/api_reference/full_reference/#albumentations.augmentations.geometric.transforms.PadIfNeeded) also supports other positioons (e.g. image in top left corner, pad on right and bottom) so I left the ability to do that in there -  should be just the case of setting the padding values depending on the parameter. Might want to test it out and add support for this as well before merge.

src/luxonis_ml/loader/augmentations.py
klemen1999: Yeah, `mask_value` is set to 0 by default

src/luxonis_ml/loader/augmentations.py
```diff
@@ -11,44 +12,36 @@ def __init__(self, train_rgb: bool = True):
         """Base class for augmentations that are used in LuxonisLoader
 
         Args:
-            train_rgb (bool, optional): Flag if should use RGB or BGR images. Defaults to True.
+            train_rgb (bool, optional): Weather should use RGB or BGR images. Defaults to True.
```
conorsim, line 15:
```suggestion
            train_rgb (bool, optional): Whether should use RGB or BGR images. Defaults to True.
```

src/luxonis_ml/loader/augmentations.py
```diff
@@ -11,44 +12,36 @@ def __init__(self, train_rgb: bool = True):
...
+    ) -> Tuple[A.BatchCompose, A.Compose]:
         """Parses provided config and returns Albumentations BatchedCompose object and Compose object for default transforms
 
         Args:
-            image_size (list): Desired image size [H,W]
-            augmentations (list): List of augmentations to use and their params
-            keep_aspect_ratio (bool, optional): Flat if should use resize that keeps aspect ratio of original image. Defaults to True.
+            image_size (List[int]): Desired image size [H,W]
+            augmentations (List[Dict[str, Any]]): List of augmentations to use and their params
+            keep_aspect_ratio (bool, optional): Weather should use resize that keeps aspect ratio of original image. Defaults to True.
```
conorsim, line 34:
```suggestion
            keep_aspect_ratio (bool, optional): Whether should use resize that keeps aspect ratio of original image. Defaults to True.
```

src/luxonis_ml/loader/augmentations.py
```diff
@@ -334,4 +389,197 @@ def __init__(
...
+        self,
+        height: int,
+        width: int,
+        interpolation: int = cv2.INTER_LINEAR,
+        border_value: int = 0,
+        mask_value: int = 0,
+        always_apply: bool = False,
+        p: float = 1.0,
+    ):
+        """Augmentation to apply letterbox resizing to images. Alse transforms
```
conorsim, line 414:
```suggestion
        """Augmentation to apply letterbox resizing to images. Also transforms
```

conorsim (approved):
LGTM - just some typos

PR #28, opened 2023-10-13
fix LDF construction if bbox list is empty

jkbmrz:
Moving this PR from main to dev. It's a fix for constructing LDF with data instances without bbox annotations. Currently, only way to add a data instance without bbox annotation is to make an addition without 'boxes' key. However, this causes troubles during training. So we either have to fix it here or in luxonis/models. I'm suggesting a PR here because a fix is quite easy and short.


conorsim (approved):
LGTM

klemen1999 (approved):

PR #32, opened 2023-11-01
Parquet Data Initial Refactor

conorsim:
Initial MVP refactor replacing the dependency on `fiftyone` with using parquet files. This is only an implementation of "offline" mode, which uses 
- See `examples/COCO_people_subset.ipynb` for an example workflow using the new `LuxonisDataset`
- Can integrate with the `data_refactor` branch of `models`

Notes:
- Versioning, views, and querying are no longer supported in offline mode
- `LuxonisDataset` no longer needs to be initialized with a `with` statement
- Users in offline mode should only track `dataset_name` when working with datasets
- Also started a `luxonis_ml/enums` for enums that are shared across the ML ops stack

Next steps not included in this PR:
- Support for cloud buckets
- Annotation deletion
- Parent and children annotations
- Test cases

I'd be particularly interested in us evaluating if the data model for annotations is as future-proof as possible.


src/luxonis_ml/data/utils/parquet.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def _get_current_parquet_file(self) -> str:
+        """Finds the best parquet file to edit based on the file size and most last write time"""
+
+        path = self._generate_filename(self.num)[1]
+        current_size = os.path.getsize(path) / (1024 * 1024)
+        if current_size < self.file_size:
+            return path
+        else:
+            for file in self.files:
```
tersekmatija, line 50: How does this look like in practice? Can some random file be the smallest?

src/luxonis_ml/data/utils/enums.py
```diff
@@ -0,0 +1,59 @@
...
+    KEYPOINTS = "keypoints"  # keypoint skeleton instances
+    LABEL = "label"  # an arbitrary label of string, bool, or number
+
+
+class MediaType(Enum):
+    """Individual file type"""
+
+    IMAGE = "image"
+    VIDEO = "video"
+    POINTCLOUD = "point cloud"
```
tersekmatija, line 25: Let's try to avoid " ".

src/luxonis_ml/data/utils/enums.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class ImageType(Enum):
+    """Image type for IMAGE HType"""
+
+    COLOR = "color"
+    MONO = "mono"
+    DISPARITY = "disparity"
```
tersekmatija, line 33: Do we need to differentiate between disparity/depth if we do have mono?

src/luxonis_ml/data/utils/constants.py
```diff
@@ -0,0 +1,17 @@
+from luxonis_ml.enums import LabelType
```
tersekmatija, line 1: Is there a reason we import both enums differently?

src/luxonis_ml/data/utils/constants.py
```diff
@@ -0,0 +1,17 @@
...
+from .enums import DataLabelType
+from typing import List, Dict, Union
+
+LDF_VERSION = "1.0.0"
+
+ANNOTATIONS_SCHEMA = {
+    "file": str,  # path to file on local disk or object storage
+    "class": str,  # any string specifying what the annotation is
```
tersekmatija, line 9: Shouldn't "class" and "type" be enums?

src/luxonis_ml/data/utils/constants.py
tersekmatija: Ah I see now. Should we move them all to the same place?

src/luxonis_ml/data/dataset.py
```diff
@@ -218,141 +129,96 @@ def __init__(
         if not isinstance(self.bucket_storage, BucketStorage):
             raise Exception(f"Must use a valid BucketStorage!")
 
+        self.online = self.team_id != "offline"
```
tersekmatija, line 132: Should we make offline mode more clear compared to setting it as a team_id.

src/luxonis_ml/data/utils/enums.py
conorsim: You mean a space right?

src/luxonis_ml/data/utils/enums.py
```diff
@@ -0,0 +1,59 @@
...
+        a BOX has a POLYLINE instance segmentation,
+        etc.)"""
+
+    CLASSIFICATION = (
+        "classification"  # used for single, multi-class, or multi-label classification
+    )
+    BOX = "box"  # bounding box
+    POLYLINE = "polyline"  # polyline to represent segmentation mask instances
+    KEYPOINTS = "keypoints"  # keypoint skeleton instances
+    LABEL = "label"  # an arbitrary label of string, bool, or number
```
klemen1999, line 17: Could we add a suppord for 3D array like label type that can be then used for for example heatmap labels and similar 2D based image-like labels? I would have it as 3D so it also supports multiple "masks" (e.g. heatmap for each keypoint class). And I would put as a separate type (opposed to merging it to LABEL) so it's easier to manage for augmentations - these would go through image augmentations (e.g. rotate, crop,..) where as LABEL wouldn't as those are "1D types"

klemen1999 (commented):
LGTM otherwise, really like the move away from `with...` statement which will allow easier swapping to custom, user defined datasets/loaders

src/luxonis_ml/data/utils/enums.py
conorsim: I would say yes since these images would probably be uint16. Although I'm not sure if we need depth.

src/luxonis_ml/data/utils/constants.py
conorsim: I was thinking the ones in the `data` module could just be used within the `data` module while the ones in `luxonis_ml/enums` can be used across the whole MLops stack. If you think it's cleaner to put them all in the same place I think that's fine.

src/luxonis_ml/data/utils/constants.py
conorsim: At first I had `type` as `Union[str, LabelType]`. But I thought it's easier from the user perspective to not import the enums and just use their string values instead. But if we want to force enum use or provide either option it's fine.

For `class` it should definitely be a string since it's user defined (e.g. person, toothbrush, cat, dog). It's the literal class of the annotation.

src/luxonis_ml/data/dataset.py
conorsim: Yeah I thought about this as well. Maybe setting an environment variable and/or a `config.json` for `data` would make sense? Like `LUXONISML_ONLINE=true`?

src/luxonis_ml/data/utils/enums.py
conorsim: This could also apply to segmentation maps as well (as opposed to polyline), right? I like the idea, but it might be interesting to test whether we should store these in parquet files or numpy array (or the like) with just paths to these arrays in annotation `value`.

I will add it as an upcoming feature, thanks!

conorsim:
Updates:
- Merging with `submodule_refactor`
- New functions in `LuxonisFileSystem` for `data`
- Initial support using GCS bucket storage tested

src/luxonis_ml/utils/filesystem.py
```diff
@@ -135,11 +222,58 @@ def read_to_byte_buffer(self) -> BytesIO:
...
+                download_path = os.path.join(self.path, remote_path)
+            else:
+                download_path = self.path
+            with self.fs.open(download_path, "rb") as f:
                 buffer = BytesIO(f.read())
 
         return buffer
 
+    def get_file_uuid(self, path: str, local: bool = False) -> str:
+        """Reads a file and returns Byte buffer
```
klemen1999, line 235: This description is probably wrong

src/luxonis_ml/data/loader.py
```diff
@@ -1,22 +1,15 @@
...
 from pathlib import Path
-from fiftyone import ViewField as F
-
-
-class LabelType(str, Enum):
-    CLASSIFICATION = "class"
-    SEGMENTATION = "segmentation"
-    BOUNDINGBOX = "boxes"
-    KEYPOINT = "keypoints"
+from luxonis_ml.enums import LabelType
```
klemen1999, line 12: Will also require update on `models` side as well

klemen1999 (approved):
Other than a small nitpick LGTM

src/luxonis_ml/utils/filesystem.py
conorsim: Thanks, will push a quick fix

src/luxonis_ml/data/utils/enums.py
tersekmatija: Yes

src/luxonis_ml/data/dataset.py
```diff
@@ -196,16 +102,22 @@ def __init__(
...
         )
+        credentials_cache_file = str(Path(self.base_path) / "credentials.json")
+        if os.path.exists(credentials_cache_file):
+            with open(credentials_cache_file) as file:
+                self.config = json.load(file)
+        else:
+            self.config = {}
+
+        if team_id is None:
+            team_id = self._get_config("LUXONISML_TEAM_ID", "offline")
```
tersekmatija, line 116: Given both are set to _ID, is there still a reason why we need both?

src/luxonis_ml/data/dataset.py
```diff
@@ -196,16 +102,22 @@ def __init__(
...
+        if team_name is None:
+            team_name = self._get_config("LUXONISML_TEAM_ID", "offline")
+
+        self.bucket = self._get_config("LUXONISML_BUCKET", None)
```
tersekmatija, line 120: Does this come with the provider as well? (GCS, S3, ...)

src/luxonis_ml/data/dataset.py
```diff
@@ -218,175 +130,107 @@ def __init__(
...
-
-        if len(res):
-            assert len(res) == 1
-            self.dataset_name = res[0]["dataset_name"]
-            self.dataset_doc = fop.LuxonisDatasetDocument.objects.get(
-                team_id=self.team_id, dataset_name=self.dataset_name
-            )
-            self.dataset_id = str(self.dataset_doc.id)
-            self.full_name = self.full_name = f"{self.team_id}-{self.dataset_id}"
+        if self.bucket_storage.value != "local" and self.bucket is None:
```
tersekmatija, line 133: Should we use enums instead of strings?

src/luxonis_ml/data/dataset.py
```diff
@@ -397,162 +241,97 @@ def _init_path(self) -> None:
...
+                / self.dataset_name
             )
             os.makedirs(self.path, exist_ok=True)
+            self.media_path = os.path.join(self.path, "media")
+            self.annotations_path = os.path.join(self.path, "annotations")
+            self.metadata_path = os.path.join(self.path, "metadata")
+            os.makedirs(self.media_path, exist_ok=True)
+            os.makedirs(self.annotations_path, exist_ok=True)
+            os.makedirs(self.metadata_path, exist_ok=True)
         elif self.bucket_storage.value == "s3":
```
tersekmatija, line 253: Given the path is the same should we just set the prefix to "GCS" and "S3" instead of having both?

tersekmatija: Instead of them being separate*

src/luxonis_ml/data/dataset.py
```diff
@@ -397,162 +241,97 @@ def _init_path(self) -> None:
...
-                    instance_ids.add(iid["instance_id"])
-            instance_ids = list(instance_ids)
+    def _get_file_index(self) -> Optional[pd.DataFrame]:
+        index = None
+        if self.bucket_storage.value == "local":
+            file_index_path = os.path.join(self.metadata_path, "file_index.parquet")
+        else:
+            file_index_path = os.path.join(".luxonisml_tmp", "file_index.parquet")
+            try:
+                self.fs.get_file("metadata/file_index.parquet", file_index_path)
```
tersekmatija, line 292: So this would mean on public bucket we'd have a ".luxonisml_tmp" directory?

src/luxonis_ml/utils/filesystem.py
```diff
@@ -110,14 +114,95 @@ def put_file(
                 else:
                     raise KeyError("No active mlflow_instance provided.")
             else:
+                import mlflow
+
                 client = mlflow.MlflowClient(tracking_uri=self.tracking_uri)
```
tersekmatija, line 119: Remind me again why we sometimes already have a mlflow client and sometimes not?

tersekmatija (approved):

src/luxonis_ml/utils/filesystem.py
conorsim: CC @klemen1999. I moved the import statement within the conditionals to not have mlflow as a requirement in luxonis_ml/utils. But we can also just make it a requirement.

src/luxonis_ml/data/dataset.py
conorsim: The provider is currently specified when initializing `LuxonisDataset`. Do you think it would make more sense to come from an environment variable?

src/luxonis_ml/data/dataset.py
conorsim: No, it's a temporary local directory that gets removed after the `file_index.parquet` is uploaded to the bucket.

src/luxonis_ml/data/dataset.py
conorsim: Thanks, fixed.

src/luxonis_ml/data/dataset.py
conorsim: Yeah I agree, fixed

src/luxonis_ml/data/dataset.py
conorsim: Fixed

src/luxonis_ml/utils/filesystem.py
klemen1999: The reson for this is to support live runs (like saving checkpoint at the end to current run). When there is already an mlflow instance active (in this is active in tracker) I wasn't able to connect to it by creating an MLFlowClient. But I'll look into it again now that I'm more familiar with this.

PR #34, opened 2023-11-24
Changed config handler

kozlov721:
Renamed `ConfigHandler` to `Config` and made it inherit from `pydantic.BaseConfig`. This removes the necesity to define two classes (`ConfigHandler` and pydantic `Config`)  when using this util.

### DIFF
--- a/src/luxonis_ml/utils/config.py
+++ b/src/luxonis_ml/utils/config.py
@@ -41,6 +41,13 @@ class Config(BaseModel):
                 raise ValueError("Provide either config path or config dictionary.")
 
             cls.instance = super().__new__(cls)
+            cls._fs = None
+            if isinstance(cfg, str):
+                from dotenv import load_dotenv
+
+                load_dotenv()
+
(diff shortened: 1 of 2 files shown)


src/luxonis_ml/utils/config.py
```diff
@@ -70,7 +93,7 @@ def save_data(self, path: str) -> None:
             path (str): Path to output yaml file
         """
         with open(path, "w+") as f:
-            yaml.dump(self._data.model_dump(), f, default_flow_style=False)
+            yaml.dump(self.model_dump(), f, default_flow_style=False)
 
     def get(self, key_merged: str, default: Any = None) -> Any:
```
klemen1999, line 98: Is this still needed or should we remove it? Maybe a point for keeping it would be still support "safe" get() with default value if it doesn't exist (similar as python dict).

kozlov721: Yeah, the `get` is still very useful (for example `MetadataLogger` depends on it).

PR #36, opened 2023-11-30
option to use python literals in config overrides

kozlov721:
As discussed in `topology_refactor` PR, I've added an option to use quoted python literals in `Config`' override list.
This is to add back the option to override entire sections without using json format in the CLI. 

In `luxonis_train`, it is now possible to do e.g.:
```bash
python -m luxonis_train train --config configs/coco_test.yaml \ 
    train.preprocessing.train_image_size '[128,128]' \
    train.callbacks '[{"name": "LearningRateMonitor", "params": {"logging_interval": "step"}}]'
```

### DIFF
--- a/src/luxonis_ml/utils/config.py
+++ b/src/luxonis_ml/utils/config.py
@@ -1,5 +1,6 @@
 import yaml
 import warnings
+import ast
 from typing import (
     Optional,
     Union,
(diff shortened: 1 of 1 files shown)


klemen1999 (approved):
Tested it with a few more tests and LGTM

PR #37, opened 2023-11-30
Luxonis logging

kozlov721:
Added a function to globally set up logging.
Can make use of rich logging and can configure handling of warnings from `warning.warn` (for warnings from other libraries).

image

### DIFF
--- a/src/luxonis_ml/utils/logging.py
+++ b/src/luxonis_ml/utils/logging.py
@@ -1,0 +1,75 @@
+import logging
+import os
+import warnings
+from typing import Literal
+
+from rich.console import Console
+from rich.logging import RichHandler
+from rich.theme import Theme
+
+
+def setup_logging(
+    *,
+    file: str | None = None,
+    use_rich: bool = False,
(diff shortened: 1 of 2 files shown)


src/luxonis_ml/utils/logging.py
```diff
@@ -0,0 +1,75 @@
...
+        use_rich (bool, optional): If True, rich library will be used for logging.
+          Defaults to False.
+        level (str, optional): Logging level. One of "DEBUG", "INFO", "WARNING",
+          "ERROR", and "CRITICAL". Defaults to "INFO".
+          If "DEBUG" is set in environment variables, the level is changed to "DEBUG".
+        configure_warnings (bool, optional): If True, warnings will be logged.
+          Defaults to True.
+    """
+    # NOTE: So we can simply run `DEBUG= python ...`
+    if "DEBUG" in os.environ:
```
tersekmatija, line 36: Would it make sense having some dedicated variable that can set either of the levels instead?

PR #38, opened 2023-11-30
Loader remote sync

conorsim:
Initial support for syncing from bucket storage in the loader, allowing initial support for datasets stored in bucket storage.

### DIFF
--- a/src/luxonis_ml/data/dataset.py
+++ b/src/luxonis_ml/data/dataset.py
@@ -235,18 +235,20 @@ class LuxonisDataset:
     def _init_path(self) -> None:
         """Configures local path or bucket directory"""
 
+        self.local_path = str(
+            Path(self.base_path)
+            / "data"
+            / self.team_id
+            / "datasets"
+            / self.dataset_name
+        )
+        self.media_path = os.path.join(self.local_path, "media")
+        self.annotations_path = os.path.join(self.local_path, "annotations")
+        self.metadata_path = os.path.join(self.local_path, "metadata")
+
         if self.bucket_storage == BucketStorage.LOCAL:
-            self.path = str(
-                Path(self.base_path)
-                / "data"
-                / self.team_id
-                / "datasets"
-                / self.dataset_name
-            )
+            self.path = self.local_path
             os.makedirs(self.path, exist_ok=True)
-            self.media_path = os.path.join(self.path, "media")
-            self.annotations_path = os.path.join(self.path, "annotations")
-            self.metadata_path = os.path.join(self.path, "metadata")
             os.makedirs(self.media_path, exist_ok=True)
             os.makedirs(self.annotations_path, exist_ok=True)
             os.makedirs(self.metadata_path, exist_ok=True)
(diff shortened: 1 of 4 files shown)


src/luxonis_ml/data/dataset.py
```diff
@@ -167,6 +170,7 @@ def __init__(
         elif self.bucket_storage == BucketStorage.AZURE_BLOB:
             raise NotImplementedError
         else:
+            self.tmp_dir = ".luxonis_tmp"
```
tersekmatija, line 173: Would it make sense making this configurable or allowing it to be project specific? E.g. if I want to store all of the files in some data folder inside my project instead of to tmp_dir?

src/luxonis_ml/utils/requirements.txt
```diff
@@ -1,4 +1,4 @@
-pydantic>=2.4.0
+pydantic>=1.10.7
```
tersekmatija, line 1: Any reason for a whole version drop?

tersekmatija (approved):

src/luxonis_ml/data/dataset.py
conorsim: You mean if you want your local storage and remote storage to be always synced (you want to store files in both)?

src/luxonis_ml/utils/requirements.txt
conorsim: Changed back

src/luxonis_ml/data/dataset.py
tersekmatija: I meant just in general to have local storage at /home/user/Projects/mymodel/data/ for example instead of /home/user/.luxonis_tmp.

conorsim: The temp directory is only meant for remote mode as a place to stage files before syncing to the bucket. In local mode, we have the base_path. This is operating on the assumption that in remote mode you wouldn't _also_ want to store files locally.

src/luxonis_ml/data/loader.py
```diff
@@ -55,19 +57,29 @@ def __init__(
...
 
         self.dataset = dataset
+        self.stream = stream
+        self.sync_mode = (
+            self.dataset.bucket_storage != BucketStorage.LOCAL and not self.stream
+        )
 
-        if self.dataset.bucket_storage.value == "local":
+        if self.sync_mode:
+            print("Syncing from cloud...")
```
kozlov721, line 66: We can use the logger here.

src/luxonis_ml/data/dataset.py
```diff
@@ -313,6 +320,12 @@ def _end_time(self) -> None:
         self.t1 = time.time()
         print(f"Took {self.t1 - self.t0} seconds")
 
+    def _make_temp_dir(self) -> None:
```
kozlov721, line 323: Might be better to create a context manager for this. A little cleaner to read and also, we ensure the 
directory gets deleted in case some exception happens.

```python
@contextmanager
def _temp_dir(self):
    os.makedirs(self.tmp_dir, exist_ok=True)
    yield
    shutil.rmtree(self.tmp_dir)
```

kozlov721 (commented):
Left some comments, otherwise LGTM

src/luxonis_ml/data/dataset.py
conorsim: Yeah, I like the idea but I don't think it would work well with the other cases. I've added a line at initialization of `_make_temp_dir` to ensure the directory is removed to begin with in case there are some leftover files, so it would at least prevent cases where an exception causes leftover files to be wrongfully synced.

PR #39, opened 2023-11-30
Modelconverter integration

conorsim:
- Adds some functionality to `LuxonisFileSystem` for use in `modelconverter`
- Removes problematic syntax that causes issues with Python 3.8

### DIFF
--- a/src/luxonis_ml/utils/filesystem.py
+++ b/src/luxonis_ml/utils/filesystem.py
@@ -308,12 +308,7 @@ class LuxonisFileSystem:
     def split_full_path(path: str) -> Tuple[str, str]:
         """Returns a tuple for the absolute and relative path given a full path"""
 
-        if os.path.isdir(path):
-            directory = os.path.dirname(path)
-            return path.split(directory)[0], directory
-        else:
-            basename = os.path.basename(path)
-            return path.split(basename)[0], basename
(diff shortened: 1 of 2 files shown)


src/luxonis_ml/utils/filesystem.py
```diff
@@ -120,6 +115,8 @@ def put_file(
                 client.log_artifact(run_id=self.run_id, local_path=local_path)
 
         elif self.is_fsspec:
+            print(local_path)
```
kozlov721, line 118: forgotten prints

src/luxonis_ml/utils/filesystem.py
```diff
@@ -284,3 +295,41 @@ def _split_mlflow_path(self, path: str) -> List[Optional[str]]:
...
+        return os.path.split(path)
+
+    @staticmethod
+    def get_protocol(path: str) -> str:
+        """Gets the detected protocol of a path"""
+
+        return _get_protocol_and_path(path)[0]
+
+
+def _get_protocol_and_path(path: str) -> Tuple[str]:
```
kozlov721, line 323: should be `Tuple[str, str]`

kozlov721 (commented):
Left a few comments, otherwise LGTM

PR #40, opened 2023-12-01
Pydantic Settings for env variables

kozlov721:
Added a new `pydantic-settings` model that loads env variables. This way we also enforce standardized naming. 
It also handles .env files so we can drop dependency on `dotenv-python`.

The `Environ` class can be added as a section to a subclassed `Config` for the ability to specify env variables there.


tersekmatija (approved):
We can drop some variables as discussed offline, LGTM otherwise!

PR #41, opened 2023-12-05
Segmentation mask support

conorsim:
Allows segmentation mask support using paths to local files. If using remote bucket storage, this supports uploading the masks to bucket storage and loading with syncing.


src/luxonis_ml/data/dataset.py
```diff
@@ -502,6 +504,50 @@ def _add_process_batch():
                 )
                 self._end_time()
 
+            mask_paths = list(
+                set(
+                    [
+                        data["value"]
+                        for data in batch_data
+                        if data["type"] == "segmentation"
```
tersekmatija, line 512: Is there a reason for or?

src/luxonis_ml/data/dataset.py
```diff
@@ -502,6 +504,50 @@ def _add_process_batch():
...
+                    self._end_time()
+                print("Finalizing paths...")
+                for data in tqdm(batch_data):
+                    if (
+                        data["type"] == "segmentation"
+                        or data["type"] == DataLabelType.SEGMENTATION
+                    ):
+                        if self.bucket_storage != BucketStorage.LOCAL:
+                            remote_path = mask_upload_dict[data["value"]]
+                            remote_path = f"{self.fs.protocol}://{os.path.join(self.fs.path, remote_path)}"
```
tersekmatija, line 546: Could we store local paths only, and prefix is defined somewhere to differentiate between LOCAL and remote? E.g., path would always be fixed to some degree, as the structure would be the same if on remote bucket or local, no?

tersekmatija (approved):
Thanks, how would instance segmentation look like? masks_instance directory and the rest more or less similar?

src/luxonis_ml/data/dataset.py
conorsim: At the moment no, since in local mode (no bucket storage) we've stopped duplicating the data and copying files to some standard path. Instead, we load from the original provided filepaths. The idea was to make it more friendly for local development and not duplicate data. Thoughts?

conorsim: The way it's currently implemented is that we have this standardization for remote storage, so yes to the first bullet. 

But to the second we do not copy any data to a standard local path. Instead, it's up to the user to generate the masks and have them saved wherever and then provide paths to them. So you could theoretically have masks in different directories across your filesystem and it would still work.

Since we will likely switch to RLE, we could imagine the same for a voxel annotation. We would avoid copying the data locally, assuming the user already has it saved somewhere, and just point to it.

tersekmatija: Ok this makes sense, but wouldn't this make syncing a bit tricky? And if you push to bucket and pull from it, it would likely re-structure the things around no?

It still seems better than doing any copying though. Maybe a random idea and would have to be thought out more (not sure if it can work), but what about having this "standard path" per annotation type. So e.g. you could have custom path for segmentation still. It would still have to be standardized to some degree though, so not sure if worth it.

conorsim:
After the initial PR, we realized we'd prefer the RLE format representation for segmentation masks. However, we plan to also support annotation types that utilize completely unstructured saved files, so we've transferred the old logic for saved segmentation masks to array types. In sum,
- `DataLabelType.SEGMENTATION` now utilizes RLE instead of paths to external masks
- `DataLabelType.ARRAY` was introduced

src/luxonis_ml/data/dataset.py
```diff
@@ -502,6 +506,47 @@ def _add_process_batch():
...
+                        or data["type"] == DataLabelType.ARRAY
+                    ]
+                )
+            )
+            if len(array_paths):
+                print("Checking arrays...")
+                self._start_time()
+                data_utils.check_arrays(array_paths)
+                self._end_time()
+                print("Generating array UUIDs...")
```
kozlov721, line 524: Could we use the new logger instead of prints?

src/luxonis_ml/data/utils/data_utils.py
```diff
@@ -121,3 +148,37 @@ def _check_keypoints(value: Any) -> None:
...
+    if not isinstance(value, str):
+        raise Exception(f"Array value {value} must be a path to a numpy array (.npy)")
+    if not _check_valid_array(value):
+        raise Exception(f"Array at path {value} is not a valid numpy array (.npy)")
+
+
+def _check_valid_image(path: str) -> bool:
+    try:
+        image = cv2.imread(path)
+        if image is None:
```
kozlov721, line 163: `return image is not None` might be cleaner

src/luxonis_ml/data/dataset.py
conorsim: Added. Do you mind checking if it's used as intended?

klemen1999 (approved):
LGTM!

src/luxonis_ml/data/dataset.py
```diff
@@ -318,7 +324,7 @@ def _start_time(self) -> None:
 
     def _end_time(self) -> None:
         self.t1 = time.time()
-        print(f"Took {self.t1 - self.t0} seconds")
+        logging.info(f"Took {self.t1 - self.t0} seconds")
```
kozlov721, line 327: Better to use `self.logger.info(...)` instead of `logging.info(...)` (the former contains additional info).

kozlov721:
I'll write instructions on how to use the logger here instead of in the code comments. (CC @klemen1999 , @conorsim )

The `setup_logging` and `reset_logging` functions should not be called anywhere in the entire `luxonis-ml`. This is a helper module, so we cannot assume how does the caller want to log the messages (like whether to use rich or log to file). 
The functions are for the caller to configure the logging globally on its side. 

In `luxonis-ml`, usage of the logging should look like:

```python
import logging  # the built in logging, not ours 
logger = logging.getLogger(__name__)

# the rest of the code

# instead of print("info message")
logger.info("info message")

# instead of warning.warn("warning message")
logger.warning("warning message")
```

kozlov721 (approved):
LGMT

PR #43, opened 2023-12-08
Filesystem - put bytes

conorsim:
Small change adding a function to upload bytes to bucket storage. Also added `rich` to requirements since I noticed it missing from a fresh environment.


kozlov721 (approved):
LGTM

PR #44, opened 2023-12-10
Parser refactor

klemen1999:
Updated LuxonisParser to newest LDF. Goal is to have automatic conversion of common annotation formats into LDF.

Note: Will add seg_mask parser once #41  is merged


kozlov721 (approved):
LGTM

src/luxonis_ml/data/dataset.py
```diff
@@ -658,33 +651,50 @@ def make_splits(
...
                 index = self._get_file_index()
                 if index is None:
                     raise Exception("File index not found")
-                if set(definitions.keys()) != {"train", "val", "test"}:
-                    raise Exception(
-                        "Must specify train, val, and test and those keys only"
-                    )
                 for split in "train", "val", "test":
+                    if split not in definitions:
+                        continue
```
conorsim, line 664: What is the case where we wouldn't have a split here? I'm not sure if it's something we should throw an error for instead since doesn't luxonis-train expect these 3 splits?

klemen1999: The case would be where we create the dataset and then down the line we only want to update "train" split. In this case `definitions` can only be `{train:[....]}` and then only this part of the splits.json will be updated (previously definitions had to have all three keys and whole splits.json was overwritten). This then also matches with usecase of `.parse_raw_dir()` function

conorsim (approved):
Looks good

PR #45, opened 2023-12-10
Updated Config

kozlov721:
# Updated Config

## Changed at what stage are the CLI overrides applied

The overrides are now applied before the `pydantic` model is populated. That is the raw input dictionary is overriden first before
passing it to the `BaseModel` constructor. 

This fixes a situation when we have two config parameters; `foo` and `bar`, and ve have a validator that ensures that both must be provided. Then in the config file we only specify `foo`, whereas `bar` is specified as CLI override. The previous implementation would reject this, because the `BaseModel` is first populated and validated from the config file only. 

## Changed the singleton pattern

The previous implementation wasn't actually singleton. It took me some time to debug this, and it turned out the commonly used singleton pattern with overriding `__new__` doesn't actually work if the class also implements `__init__`. If implemented, `__init__` is **always** run on what's returned from `__new__`, even if it's an already initialized instance. This is hardcoded to cpython and cannot be changed.  So, in general, one should either only implement `__new__`, or only implement `__init__`. Implementing both can lead to some hard-to-find bugs.

## Added unittests and coverage

Added unittests for the `LuxonisConfig` and a GH action to run the tests on PR. The action also generates a coverage report. I'd like if we'd eventually achieved at least 90 % test coverage.


README.md
```diff
@@ -110,7 +117,7 @@ additions [list]:
         filepath [str]: local path to an image to be added
         class [str]: for Image Classification, the name of the class for the entire image
         boxes [list]: for Object Detection, a list of bounding boxes
-            box [list]: a list of length 5 encoding the following information - [class, x, y, width, height], where x, y, width, and height are all normalized to the image dimensions 
+            box [list]: a list of length 5 encoding the following information - [class, x, y, width, height], where x, y, width, and height are all normalized to the image dimensions
```
conorsim, line 120: Just FYI - this README is super out of date. Any section on LDF needs to be rewritten

conorsim (approved):
LGTM

PR #47, opened 2023-12-12
Fix/segmentation array loader

kozlov721:
# loader and classes fix
- support for "array" segmentation in loader
- support for different classes by task


examples/COCO_people_subset.ipynb
```diff
@@ -149,20 +150,23 @@
...
     "                    \"class\": \"person\",\n",
     "                    \"type\": \"polyline\",\n",
     "                    \"value\": poly\n",
     "                }\n",
     "            else: # RLE format\n",
-    "                rle = (seg[\"size\"][0], seg[\"size\"][1], seg[\"counts\"])\n",
+    "                height, width = seg[\"size\"]\n",
+    "                rle = mask_utils.frPyObjects(seg, height, width)\n",
+    "\n",
+    "                value = (rle[\"size\"][0], rle[\"size\"][1], list(rle[\"counts\"]))\n",
```
conorsim, line 164: In this example we either need to restore to what it was or not cast to `list` since the `rle["counts"]` are `bytes` here

conorsim (commented):
Just this one comment needs a quick fix, then I think good to merge!

PR #49, opened 2023-12-16
Cached LuxonisFileSystem

kozlov721:
Added option to specify cache storage to the filesystem. Unfortunately, it seems the cached fsspec works only with `fsspec.open` (`read_to_byte_buffer`), and not with `fsspec.download` (`get_file`). At least not with S3 and GCS filesystems, but still better than nothing.


src/luxonis_ml/utils/filesystem.py
```diff
@@ -72,22 +78,26 @@ def full_path(self) -> str:
...
                 secret=environ.AWS_SECRET_ACCESS_KEY,
                 endpoint_url=environ.AWS_S3_ENDPOINT_URL,
             )
         elif self.protocol == "gcs":
             # NOTE: This should automatically read from GOOGLE_APPLICATION_CREDENTIALS
-            return fsspec.filesystem(self.protocol)
+            fs = fsspec.filesystem(self.protocol)
         elif self.protocol == "file":
-            return fsspec.filesystem(self.protocol)
+            fs = fsspec.filesystem(self.protocol)
```
conorsim, line 94: Would there be some weird behavior if we use `file` protocol and also `cache_options` is not `None`?

kozlov721: I briefly tested it and it should probably work fine, but I'm now ignoring the cache settings when protocol is `"file"` to be safe.

PR #51, opened 2023-12-17
Fix/safe import

kozlov721:
Added a simple guard to raise a custom `ImportError` when attempting to import a package
without the correct extra dependencies installed.


conorsim (approved):
Nice, looks good! I briefly tested it out

PR #52, opened 2023-12-19
Added delete functions to LFS

DrejcPesjak:
Added (and tested) 3 methods to LuxonisFileSystem:
 - delete_file
 - delete_files
 - delete_dir

Motivation was the Robothub DataIngestion Airflow pipeline, where we download and filter images from RH, and need a cleanup afterwards (deleting tmp files).

### DIFF
--- a/src/luxonis_ml/utils/filesystem.py
+++ b/src/luxonis_ml/utils/filesystem.py
@@ -215,6 +215,26 @@ class LuxonisFileSystem:
             self.fs.download(
                 os.path.join(self.path, remote_path), local_path, recursive=False
             )
+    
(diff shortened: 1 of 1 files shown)


kozlov721 (changes requested):
Would be better to raise a `NotImplementedError` when `self.is_fsspec` is `False`.

conorsim (changes requested):
Looks good but I would request to quickly add `ThreadPoolExecutor` to the `delete_files` method

conorsim (approved):

kozlov721 (approved):
LGTM

PR #55, opened 2023-12-27
Mixup aug

klemen1999:
Added [MixUp](https://paperswithcode.com/method/mixup) augmentation which combines two images and their annotations into one. This alos serves as an example of a custom batched transform implemented in albumentations.

Example:
mixup


src/luxonis_ml/data/augmentations.py
```diff
@@ -696,3 +697,72 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
         target_params = super().get_params_dependent_on_targets(params)
         target_params["indices"] = list(range(self.n_tiles))
         return target_params
+
+
+@AUGMENTATIONS.register_module()
+class MixUp(A.BatchBasedTransform):
+    def __init__(
+        self,
+        alpha: float = 0.5,
```
kozlov721, line 706: Could we add on option to randomize the `alpha` as well?
The `alpha` could be either:
- a single float - in that case it would be fixed
- a tuple of two floats - bounds for a uniform distribution.

kozlov721 (approved):
Left one comment, otherwise LGTM

src/luxonis_ml/data/augmentations.py
```diff
@@ -696,3 +697,72 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+    def apply_to_image_batch(self, image_batch, image_shapes, **params):
+        image1 = image_batch[0]
+        # resize second image to size of the first one
+        image2 = cv2.resize(image_batch[1], (image_shapes[0][1], image_shapes[0][0]))
+        img_out = cv2.addWeighted(image1, self.alpha, image2, 1 - self.alpha, 0.0)
+        return [img_out]
+
+    def apply_to_mask_batch(self, mask_batch, image_shapes, **params):
+        mask1 = mask_batch[0]
+        mask2 = cv2.resize(
```
conorsim, line 741: It might be interesting to also allow the option to paste onto one another instead of resizing to not affect aspect ratios. But I think it's fine without it for now.

conorsim (approved):
Looks good!

src/luxonis_ml/data/augmentations.py
klemen1999: I thought about this as well. In that case we would also need to define how to align it (e.g. align top left corner, align centers of both images, etc.) - probably additional argument. But also in most dataset I think images are of same size (not in COCO though). This could be certanly added if it turns out that there is a use case for it.

PR #65, opened 2024-01-08
Update Data Notebooks

conorsim:
Cleaned up the existing notebook demonstrating `LuxonisDataset` and created a new notebook which provides an example of using `LuxonisParser` instead on the same data.


kozlov721 (approved):
LGTM

PR #73, opened 2024-01-31
Ensure base path directory exists

conorsim:
Fixes an issue when creating a dataset for the first time. Before the fix, the `datasets.json` file would write to a directory that does not yet exist. This fix ensures the base path always exists first.


kozlov721 (approved):
LGTM

DrejcPesjak (approved):
LGTM

PR #75, opened 2024-02-02
NN archive

conorsim:
Adding back the first iteration of NN archive to integrate with `luxonis-train`. Some changes from the initial version include

- Removing `Metadata.platform` and adding `Metadata.path` instead to point to executables
- Making heads optional for models with raw output
- Adding NV12 as a `DataType`
- Adding docstrings also to the pydantic `Field` descriptions


luxonis_ml/scheme/v1/config_building_blocks/base_models/head.py
```diff
@@ -0,0 +1,192 @@
+from abc import ABC
+from typing import List, Optional, Union
+
+from pydantic import BaseModel, Field, validator
+
+from ..enums import ObjectDetectionSubtypeYOLO
```
kozlov721, line 6: These seem like invalid imports

luxonis_ml/scheme/v1/config_building_blocks/base_models/head.py
```diff
@@ -0,0 +1,192 @@
...
+class HeadMetadataClassification(HeadMetadata):
+    """Metadata for classification head.
+
+    @type family: str
+    @ivar family: Decoding family.
+    @type is_softmax: bool
+    @ivar is_softmax: True, if output is already softmaxed.
+    """
+
+    family: str = Field("Classification", Literal=True, description="Decoding family.")
```
kozlov721, line 39: `Literal` is not an argument for `Field`. Is that supposed to go to extra JSON fields? If so, it should be included in the `json_schema_extra` argument instead:

```python
family: str = Field("Classification", description="Decoding family.", json_schema_extra={"Literal": True})
```

luxonis_ml/scheme/v1/config_building_blocks/base_models/head.py
```diff
@@ -0,0 +1,192 @@
...
+    is_softmax: bool = Field(description="True, if output is already softmaxed.")
+
+    @validator("family")
```
kozlov721, line 42: In `pydantic` V2 this should be `@field_validator("family")`

luxonis_ml/scheme/v1/config_building_blocks/base_models/head.py
conorsim: Thanks. A bunch of files got screwed up in the merge

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -0,0 +1,69 @@
...
+    @type preprocessing: PreprocessingBlock
+    @ivar preprocessing: Preprocessing steps applied to the input data.
+    """
+
+    name: str = Field(description="Name of the input layer.")
+    dtype: DataType = Field(
+        description="Data type of the input data (e.g., 'float32')."
+    )
+    input_type: InputType = Field(description="Type of input data (e.g., 'image').")
+    shape: Annotated[List[int], Field(min_length=1, max_length=5)] = Field(
```
jkbmrz, line 64: a rather aesthetic change - maybe change to `shape: List[int] = Field(min_length=1, max_length=5), description=...` to avoid writing Field twice

luxonis_ml/nn_archive/config_building_blocks/base_models/custom_base_model.py
```diff
@@ -0,0 +1,6 @@
+from pydantic import BaseModel, Extra
+
+
+class CustomBaseModel(BaseModel):
+    class Config:
```
kozlov721, line 5: The `Config` class is deprecated in `pydantic` `v2`. Instead, it should be:
```python
from pydantic import BaseModel, ConfigDict

class CustomBaseModel(BaseModel)
    model_config = ConfigDict(extra="forbid")
```

PR #76, opened 2024-02-05
Sync Classes for Remote Datasets

conorsim:
When downloading a remote dataset, this also sets the dataset classes according to the `classes.json` file. This fix allows `LuxonisLoader` to be able to find the correct classes in loading a dataset located on bucket storage.


kozlov721 (approved):
LGTM

PR #77, opened 2024-02-07
Weaviate embeddings rework

DrejcPesjak:
This rework contains:

- added Weaviate vector database support
- added an abstract class VectorDB_API
- moved from required uuid as file name, to storing image file paths alongside in the vector DB
- removed pytorch dependencies from the library (still present in examples) 
- reworked the examples jupyter notebooks

Note: currently supports only cosine similarity as a distance score, as it is nicely limited between 0 and 1.


examples/utils/torch_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    # model = models.resnet50(pretrained=True) # depricated
+    model = models.resnet50(weights=resnet.ResNet50_Weights.IMAGENET1K_V1)
+    model = nn.Sequential(
+        *list(model.children())[:-1]
+    )  # Remove the last fully connected layer
+    model.eval()
+    return model
+
+
+def load_model_resnet50() -> nn.Module:
```
kozlov721, line 26: Might be nicer to merge these two functions together:

```python
def load_model_resnet50(discard_last_layer: bool):
    ...
```

conorsim:
I'll review later but it would be nice to add type hints to all the function definitions - I noticed them occasionally but not consistently.

luxonis_ml/embeddings/methods/duplicate.py
```diff
@@ -83,7 +78,7 @@ def _plot_kde(xs, s, density, maxima, minima):
     plt.show()
 
 
-def kde_peaks(data, bandwidth="scott", plot=False):
+def kde_peaks(data: np.ndarray, bandwidth: Union[str, float] = "scott", plot: bool = False) -> tuple[np.ndarray, np.ndarray, int, float]:
```
klemen1999, line 81: AFAIK luxonis-ml has python3.8>= requirement and for python3.8 `tuple` type annotation should be switched with `Tuple` or it will throw an error

conorsim (changes requested):
Thanks - I like the processing example! 

Requesting minor change due to a typing error which breaks the import in the notebooks. And then not sure if my Weaviate issue is unique to me or not.

luxonis_ml/embeddings/methods/duplicate.py
```diff
@@ -1,79 +1,74 @@
-"""Near-duplicate Search with Qdrant.
+"""
+**Near-duplicate Search with Qdrant and Weaviate**
 
-This module provides utilities to detect and remove near-duplicate data points
-within a given set of embeddings. The removal process uses Kernel Density
-Estimation (KDE) on embeddings cosine similarity for optimal split, making
-this approach particularly suited for embeddings in high dimensional spaces.
+**Overview:**
```
kozlov721, line 4: Use `B{Overview:}` for bold text instead of asterisks, see [epytext documentation](https://epydoc.sourceforge.net/epytext.html) for more details.

luxonis_ml/embeddings/methods/duplicate.py
```diff
@@ -1,79 +1,74 @@
...
+   # Based on average of multiple embeddings:
+   dark_ix = np.array([10, 123, 333, 405])
+   emb_dark = X[dark_ix]
+   remove_dark_ix, paths = find_similar_qdrant(emb_dark, qdrant_api, "image_path", 25, 5000, "average", "kde_basic", "scott", plot=True)
+   ```
+
+**Additional Notes:**
 
+- For Weaviate-specific examples, refer to the provided code examples.
+- The `search_vectordb` function can be used with either Qdrant or Weaviate, depending on the provided `vectordb_api` object.
```
kozlov721, line 60: Similarly, `C{search_vectordb}` instead of backticks.

luxonis_ml/embeddings/methods/duplicate.py
kozlov721: In addition to this, it would be nice to create at least a very dummy test case for each added module that at least imports everything, so it catches errors like this.

DrejcPesjak:
### Pre-commit hooks passed, but pre-commit CI test still fails

Ran the pre-commits as described [here](https://github.com/luxonis/luxonis-ml/blob/dev/CONTRIBUTING.md). And got the following output when commiting:

`git commit -m "Pre-commit hook changes"`

> ruff.....................................................................Passed
> ruff-format..............................................................Passed
> docformatter.............................................................Passed
> don't commit to branch...................................................Passed
> mdformat.............................................(no files to check)Skipped
> [weaviate_emb 25427a1] Pre-commit hook changes
>  5 files changed, 267 insertions(+), 235 deletions(-)

But the _CI/ call-reusable-workflow/pre-commit (pull_request)_ test still fails.

kozlov721:
The `pre-commit` command does not pass on my end with the same errors as the ones in the CI. I see where's the issue probably and will update the CONTRIBUTING.md. When commiting it only runs on the commited files, which might pass, but it doesn't mean it passes on all files. This would be caused by previously pushing without the pre-commit checks, so the affected files are upstream and no longer checked when `git commit` is ran. Run `pre-commit run --all-files` to get the errors on all files.

conorsim (approved):
Looks good now - the import issue is fixed

PR #79, opened 2024-02-29
Dataset Recognition

kozlov721:
# Dataset Recognition

- Split parsers into separate classes
  - The parsers are more low-level than before, the high-level usage is provided by the new `LuxonisParaser` class 
- Added `validate` and `validate_split` methods for use with dataset recognition
- Added a new `LuxonisParser` class that automatically recognizes dataset type of given dataset directory and parses its content using the correct parser
- Added test cases for all the supported parsers

## Example Usage

```python
parser = LuxonisParser(
    "gs://luxonis-test-bucket/luxonis-ml-test-data/COCO_people_subset",
    dataset_name="coco_test",
    delete_existing=True,
)
dataset = parser.parse()
```


examples/Data_Parser_Example.ipynb
conorsim, line 1: Not sure if it's an issue with my local environment but when I run the last cell with the loader I get
```
KeyError: <LabelType.BOUNDINGBOX: 'boxes'>
```

conorsim (approved):
Looks great from a code perspective. We'll wait to fix the workflow issue for the tests

examples/Data_Parser_Example.ipynb
kozlov721: You might need to first delete the `data` directory if it already exists. I changed some things, but it won't re-download if the zip exists.

PR #82, opened 2024-03-06
NN Archive Structural Update

conorsim:
This introduces the following changes to the NN archive structure from some iteration on what is needed for DepthAI:
- We introduce a new `HeadMetadata` for YOLO instance segmentation - `HeadMetadataInstanceSegmentationYOLO`.
- We drop the `stages` block and instead replace it with `model` since any multi-executable NN cases will instead be handled by `Head.metadata`. For example, `HeadMetadataInstanceSegmentationYOLO` specifies `postprocessor_path` for the postprocessing step which uses a second blob to perform a multiplication operation.
- We drop the `connections` block since we no longer have `stages`.
- We remove the `head_id` field in the `outputs` block and instead introduce a `Head.outputs` block which can be more descriptive about the parameters required for a DepthAI parser.
- We remove the `stride` parameter for object detection heads.
- We rework the structure of `anchors` for object detection heads to be a triply nested list. We assume it's always in the order of smallest to largest outputs.


tersekmatija:
Would `postprocessor_path` be null if we use ONNX or some other format where we don't need to have 2 "executables"?

jkbmrz (approved):
LGTM

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
```diff
@@ -172,17 +211,20 @@ def __init__(self):
...
-    @ivar head_id: Unique head identifier.
+    @type outputs: C{Union[List[str], Dict[str, Union[str, List]]]}
+    @ivar outputs: A list of output names from the `outputs` block of the archive or a dictionary mapping DepthAI parser names needed for the head to output names. The referenced outputs will be used by the DepthAI parser.
     @type metadata: HeadMetadata
     @ivar metadata: Parameters required by head to run postprocessing.
     """
 
-    head_id: str = Field(description="Unique head identifier.")
+    outputs: Union[List[str], Dict[str, Union[str, List]]] = Field(
+        description="A list of output names from the `outputs` block of the archive or a dictionary mapping DepthAI parser names needed for the head to output names. The referenced outputs will be used by the DepthAI parser."
```
jkbmrz, line 221: I assume only the outputs relevant to the head operations must be listed here? From the perspective of automatic NN archive generation, I'm wondering how do we plan to identify those. We could try to draw it from cfg model architecture but this might not be so straight-forwardas as outputs now became the outputs of the executable (e.g. ONNX) and might differ from what we have in cfg (as far as I know).

conorsim: @jkbmrz I was considering overriding `outputs` for each `HeadMetadata` or at least providing some validation on the expected format of `outputs` for a given head. However, we would then need to define `outputs` on the `HeadMetadata` level (instead of `Head`) or have one really long validtor in `Head` (which I don't think is a good idea). If we move `outputs` to `HeadMetadata`, then we only have `HeadMetadata` attributes and no attributes specific to `Head`. 

Thus, it might make sense removing `HeadMetadata` entirely and just having all the head types inherit from `Head`. Then, we overrride `outputs` for each head, further defining its type and validating possible param names (in the `outputs` keys). Thoughts on that idea?

tersekmatija:
@conorsim Let's always store JSONs as nnarchive.json or config.json to keep it consistent as discussed offline?

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
jkbmrz: We could remove the `HeadMetadata` class and have all the head types inherit from `Head` but I'm still not sure how we would figure out automatically what outputs belong to a specific head. Or does overriding `outputs` for each head means  predefining?

jkbmrz:
^ a small commit adding return statement to `ArchiveGenerator.make_archive()` method.

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
conorsim: Commit [2dcc2c3](https://github.com/luxonis/luxonis-ml/pull/82/commits/2dcc2c3db3253e289588603af1e65faeadbb5754) implements the idea

jkbmrz (approved):

jkbmrz:
^ added support for LZMA compression and setting .tar.xz as the default nn archive option

kozlov721 (approved):
LGTM

PR #83, opened 2024-03-12
Auto-Installing Protocol-Specific Packages

kozlov721:
Protocol-specific packages (`s3fs`, `gcsfs` or `mlflow`) are not installed automatically from `requirements.txt` anymore, but rather on demand after the first usage of the protocol.

This is to avoid issues when installing `luxonis-ml` on docker containers with Alpine Linux as a base.


luxonis_ml/utils/requirements.txt
```diff
@@ -3,6 +3,6 @@ pydantic-settings>=2.1.0
 PyYAML>=6.0
 fsspec>=2023.1.0
 rich>=13.6.0
-s3fs>=2023.1.0
-gcsfs>=2023.1.0
-mlflow>=2.10.0
+# s3fs>=2023.1.0
+# gcsfs>=2023.1.0
+# mlflow>=2.10.0
```
conorsim, line 8: Should we remove them entirely from here instead of having them commented?

conorsim (approved):
LGTM and seems to fix my issue with alpine

luxonis_ml/utils/requirements.txt
kozlov721: I was thinking it could stay there to document the packages we're installing, but I'm not against removing it altogether.

kozlov721: I'll at least add a comment to it

Issue #69, opened 2024-01-11
BaseDataset


conorsim:
# BaseDataset Class

In the current implementation, we only define a single `LuxonisDataset` class for dataset management. However, we wish to support using additional datasets with the `Registry`. More specifically, we wish to allow plugins from other libraries such that other libraries can define their own e.g. `PluginDataset` to override the necessary dataset functions.

## Proposed Solution

We should implement a `BaseDataset` abstract class to define the necessary methods for implementing and integrating custom dataset classes into our MLOPs stack.

### Current Method
```python
class LuxonisDataset:
    def set_classes(self, classes: List[str], ...) -> None:
        """docstring"""
        ...
```
### Proposed Method
```python
class BaseDataset(ABC, ...):
    def set_classes(self, classes: List[str], ...) -> None:
        """docstring"""
        ...

class LuxonisDataset(BaseDataset, ...):
    def set_classes(self, classes: List[str], ...) -> None:
        # override method here!
```
Opening up the possibility to define
```python
class PluginDataset(BaseDataset, ...):
    def set_classes(self, classes: List[str], ...) -> None:
        # override method here!
```

conorsim:
Resolved in #84

PR #84, opened 2024-03-13
Base Dataset and Dataset Registry

conorsim:
This PR provides an implementation of `BaseDataset` as described in https://github.com/luxonis/luxonis-ml/issues/69.

It also includes some of the following changes:
- The `dataset.py` file is now broken into its own `datasets` module for better organization
- A change to `LuxonisFileSystem.get_dir` is made so it can either be passed a list of remote paths or a path to a directory. (Similar to the way `LuxonisFileSystem.put_dir` is implemented)
- Removal of `self.online` attribute from `LuxonisDataset`
- Support for `LuxonisParser` to use a custom dataset plugin


conorsim:
Python 3.8 and 3.9 tests are failing. I only tested 3.10 locally - will figure out what the issue is.

luxonis_ml/data/datasets/base_dataset.py
```diff
@@ -0,0 +1,131 @@
...
+
+from .source import LuxonisSource
+
+DATASETS_REGISTRY = Registry(name="datasets")
+
+DatasetGenerator = Generator[Dict[str, Any], None, None]
+DatasetGeneratorFunction = Callable[[], DatasetGenerator]
+
+
+class BaseDataset(ABC, metaclass=AutoRegisterMeta, register=False):
```
kozlov721, line 14: the `registry=DATASETS_REGISTRY` argument can be also specified here in the base instead of later in `LuxonisDataset`.

kozlov721 (commented):
LGTM

PR #85, opened 2024-03-14
NN Archive Structural Fixes for YOLO Heads

conorsim:
The following changes are made to NN archive:
- `HeadKeypointDetectionYOLO` not inherits from `HeadObjectDetectionYOLO` so we can support all the normal object detection parameters, such as anchors
- `n_keypoints` is now a required parameter under `HeadKeypointDetectionYOLO`
- `n_prototypes` is now a required parameter under `HeadInstanceSegmentationYOLO` only
- `prototype_output_name` is removed since this should be handled already by #82


kozlov721 (approved):
LGTM

jkbmrz (approved):
LGTM

PR #90, opened 2024-03-27
Put File Plugin

conorsim:
This PR adds the ability to override the `LuxonisFileSystem.put_file` method with a custom implementation. Here is an example of its use:

In your custom plugin code:
```
from luxonis_ml.utils import PUT_FILE_REGISTRY

@PUT_FILE_REGISTRY.register_module()
def my_put_file(
    local_path: PathType,
    remote_path: PathType,
    mlflow_instance: Optional[ModuleType] = None,
):
   ... your custom logic here ...
```

And then in initializing LFS:
```
from my_plugin_module import my_put_file

lfs = LuxonisFileSystem(
    "s3://my/bucket/path",
    put_file_plugin=my_put_file
)

lfs.put_file(...) # call your plugin function
```

### DIFF
--- a/luxonis_ml/utils/filesystem.py
+++ b/luxonis_ml/utils/filesystem.py
@@ -15,10 +15,13 @@ from typing import Dict, Iterator, List, Optional, Sequence, Tuple, Union, cast
 import fsspec
 
 from .environ import environ
+from .registry import Registry
 
 logger = getLogger(__name__)
 
 PathType = Union[str, Path]
+
+PUT_FILE_REGISTRY = Registry(name="put_file")
 
 
 class FSType(Enum):
@@ -33,6 +36,7 @@ class LuxonisFileSystem:
         allow_active_mlflow_run: Optional[bool] = False,
         allow_local: Optional[bool] = True,
         cache_storage: Optional[str] = None,
+        put_file_plugin: Optional[str] = None,
     ):
         """Abstraction over remote and local sources.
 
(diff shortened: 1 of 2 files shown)


luxonis_ml/utils/filesystem.py
```diff
@@ -98,6 +105,9 @@ def __init__(
             self.fs = self.init_fsspec_filesystem()
         self.path = PurePosixPath(cast(str, _path))
 
+        if put_file_plugin:
+            self.put_file = PUT_FILE_REGISTRY.get(put_file_plugin)
```
kozlov721, line 109: Not sure if this shouldn't be something like this instead:

```python
custom_put_file = PUT_FILE_REGISTRY.get(put_file_plugin)
self.put_file = lambda self, *args, **kwargs: custom_put_file(*args, **kwargs)
```
to deal with the `self` argument. Maybe Python deals with it in the background though. 

Or we can require it in the signature of the plugin function:

```python
import inspect

custom_put_file = PUT_FILE_REGISTRY.get(put_file_plugin)
if inspect.getfullargspec(custom_put_file).args[0] != "self":  # or maybe "fs" instead of "self"
    raise ValueError(...)
```

kozlov721:
>And then in initializing LFS:

Shouldn't it be `put_file_plugin="my_put_file"` in the code snippet?

Also, I'm thinking if the registry is even necessary when we have to pass the name anyway. Instead of the name we can just pass the function itself instead of retrieving it from the registry by name.

luxonis_ml/utils/filesystem.py
conorsim: > Maybe Python deals with it in the background though.

It seemed like Python handled it in the background since I was wondering the same thing but I could experiment with this.

kozlov721 (approved):

PR #100, opened 2024-04-05
Fix/coco parser splits

sokovninn:
The PR includes:

- Fixed the order of val and test in the BaseParser.
- Fixed overriding test images with an empty list in the COCOParser.


kozlov721 (approved):
LGTM

conorsim (approved):

PR #102, opened 2024-04-13
Fix Remote Dataset

kozlov721:
Fixes issues in this [ClickUp task](https://app.clickup.com/t/86by8hf7x).

- support for `__len__` with remote datasets
- raising an error in case of missing `GOOGLE_APPLICATION_CREDENTIALS`

### DIFF
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -130,7 +130,9 @@ class LuxonisDataset(BaseDataset):
     def identifier(self) -> str:
         if self.dataset_name is not None:
             return self.dataset_name
-        assert self.dataset_id is not None
+        assert (
+            self.dataset_id is not None
+        ), "At least one of dataset_name or dataset_id must be provided."
         return self.dataset_id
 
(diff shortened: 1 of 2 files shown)


luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -225,6 +225,8 @@ def _load_df_offline(self, sync_mode: bool = False) -> Optional[pd.DataFrame]:
             annotations_path = self.annotations_path
         else:
             annotations_path = osp.join(self.tmp_dir, "annotations")
+        if not osp.exists(annotations_path):
+            return None
```
conorsim, line 229: Should we `return 0` here? Or if not change the return type to `Optional[int]`

kozlov721: Not sure what you mean. The function is annotated as `Optional[pd.DataFrame]`.

conorsim: My mistake - nevermind

conorsim (approved):

PR #103, opened 2024-04-15
Redefine Constraint on Duplicate Filenames

jkbmrz:
**Description**:
Currently in LuxonisDataset, we have a constraint that the base filename of a path to media being added to a dataset must be unique. However, in cases like the SOLO format, we naturally have many different images with the same filenames. We redefine here this constraint to decide duplication based on the original_filepath instead of the filename.


jkbmrz:
Done with testing parsers. Should we merge @kozlov721 @conorsim?

kozlov721 (approved):

conorsim (approved):
Yes, feel free to merge @jkbmrz

PR #107, opened 2024-05-07
Feature/skeleton loading

JSabadin:
### Title
Add Skeleton Synchronization to Fix Broadcasting Error in Luxonis ML

### Problem
Encountered a `ValueError` due to an inability to broadcast an input array from shape (52,) into shape (1,) when assigning values to `points[0, :nk + 1] = kps`.

### Solution
Implemented skeleton synchronization within the `luxonis_ml` feature on the `feature/skeleton-loading` branch to correct the shape mismatch.

### Details of Implementation
- Modified `loader.py` and `luxonis_dataset.py` to include skeleton synchronization.
- Created a new JSON file, `skeleton.json`, to store skeleton data similarly to how `classes.json` handles class data.

### Testing
- Verified locally using cached files to ensure backward compatibility and correct operation.
- Conducted remote testing to confirm that the fix addresses the failure experienced in environments without cached data.

### Additional Notes
- The implementation of skeleton synchronization mirrors the existing approach used for class data handling, promoting consistency in data management.


kozlov721 (approved):
LGTM

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -404,15 +413,22 @@ def get_classes(
...
         keypoints.
 
         @rtype: Dict[str, Dict]
         @return: A dictionary mapping classes to their skeleton definitions.
         """
+        if sync_mode:
+            local_file = osp.join(self.metadata_path, "skeletons.json")
+            self.fs.get_file("metadata/skeletons.json", local_file)
+            with open(local_file) as file:
+                skeletons = json.load(file)
```
tersekmatija, line 427: Should we be returning some warning there if there is no file?

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -404,15 +413,22 @@ def get_classes(
...
+        else:
+            skeletons = self.datasets[self.dataset_name]["skeletons"]
```
tersekmatija, line 429: Could this be empty?
Is it valid load if it's empty or should we also return a warning?

luxonis_ml/data/loader.py
```diff
@@ -91,6 +91,10 @@ def __init__(
                 synced_classes = json.load(file)
             for task in synced_classes:
                 self.dataset.set_classes(classes=synced_classes[task], task=task)
+            skeletons_file = os.path.join(self.dataset.metadata_path, "skeletons.json")
+            with open(skeletons_file) as file:
```
tersekmatija, line 95: Same comment here. I guess this will just fail if there is not skeletons.json file?

Can this affect training for other non-skeleton tasks?

tersekmatija (commented):
Thanks, added some quick questions. Can you take a look at this as well @klemen1999?

luxonis_ml/data/datasets/luxonis_dataset.py
JSabadin: Yes, it would be beneficial to include a warning here.

luxonis_ml/data/loader.py
JSabadin: Yes, it seems like it will impact other trainings. Maybe it would be wise to check if the file exists and, if so, call the get_skeletons() function?

luxonis_ml/data/datasets/luxonis_dataset.py
JSabadin: @tersekmatija I've issued a warning here. Could you please check this fix?

luxonis_ml/data/loader.py
JSabadin: @tersekmatija Could you please check this fix?

conorsim (approved):
LGTM

luxonis_ml/data/loader.py
```diff
@@ -91,6 +91,13 @@ def __init__(
                 synced_classes = json.load(file)
             for task in synced_classes:
                 self.dataset.set_classes(classes=synced_classes[task], task=task)
+            skeletons_file = os.path.join(self.dataset.metadata_path, "skeletons.json")
+            try:
+                with open(skeletons_file, "r") as file:
+                    synced_skeletons = json.load(file)
+                self.dataset.set_skeletons(synced_skeletons)
+            except FileNotFoundError:
```
klemen1999, line 99: If there is an existing remote dataset which doesn't have this file (made before these changes) will the loading still work or will it hit this error?

klemen1999 (commented):
Otherwise LGTM

luxonis_ml/data/loader.py
JSabadin: After testing the loading of the detection COCO validation dataset from the _mlcloud-dev_ _luxonis_train_project_ example, it initially failed because `self.fs.get_file("metadata/skeletons.json", local_file)` does not function without a specified file. However, I resolved this issue by moving the statement to a point after verifying the file's existence. Currently, I'm running the easyml remote command for this COCO detection example  to ensure it functions correctly. It's still in the pending state, but I will keep you updated on its progress.

PR #108, opened 2024-05-18
Add OBB head and output types

HonzaCuhel:
This PR includes adding OBB head and head output to NNArchive


conorsim (approved):
I don't have the domain knowledge on what the OBB output/model looks like but assuming that is good, LGTM

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
```diff
@@ -254,6 +255,30 @@ def validate_label_type(
...
+    @ivar family: Decoding family.
+    @type outputs: C{OutputsOBBDetectionYOLO}
+    @ivar outputs: A configuration specifying which output names from the `outputs` block of the archive are fed into the head.
+    """
+
+    family: Literal["OBBDetectionYOLO"] = Field(..., description="Decoding family.")
+    outputs: OutputsOBBDetectionYOLO = Field(
+        description="A configuration specifying which output names from the `outputs` block of the archive are fed into the head."
+    )
+
```
tersekmatija, line 271: Do we need to store any additional information here? Like number of channels...?

I see later angles are defined as a separate output, so it might not be needed. Just thinking if this might be a useful information to have here?

tersekmatija (approved):
Maybe a general question - should we keep Yolo heads separated for each task, or would having all in a single class with optional fields be better? Right now we make many heads, where each inherits `HeadObjectDetectionYOLO`, but defines a new family. In theory, it should be very easy to tell the additional task ("keypoint detection", "obb"), based on a flag that could be present on `HeadObjectDetectionYOLO` itself.

This might simplify the code, but could be more challenging if we need to handle different output objects.
 
CC @conorsim

kozlov721 (approved):
LGTM

jkbmrz (approved):
LGTM, sorry for a delay!

PR #109, opened 2024-05-27
NN Archive - Layout Info

jkbmrz:
Currently, there is no way to figure out automatically if the model requests input is grayscale. This PR adds a `model.inputs.image_layout` field that in combination with the `model.inputs.shape` field allows to determine the required input format.

### DIFF
--- a/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
+++ b/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
@@ -83,13 +83,15 @@ class Input(CustomBaseModel):
     ):
         if "layout" in values.keys():
             values["layout"] = values["layout"].upper()
+            idx_c = values["layout"].find("C")
+            idx_n = values["layout"].find("N")
 
             if len(values["layout"]) != len(values["shape"]):
                 raise ValueError("Layout and shape must have the same length.")
 
-            if values["layout"][0] != "N":
+            if idx_n not in [-1, 0]:
                 raise ValueError(
-                    "First letter of layout must always be N (batch size)."
+                    "N (batch size) must be the first letter if included."
                 )
 
             if values["input_type"] == InputType.IMAGE.value:
@@ -97,7 +99,7 @@ class Input(CustomBaseModel):
                     raise ValueError(
                         "C letter must be present in layout for image input type."
                     )
-                idx_c = values["layout"].find("C")
+                
                 if values["shape"][idx_c] not in [1, 3]:
                     raise ValueError(
                         "Color dimension for image input type must either be 1 (grayscale) or 3 (color)."
--- a/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
+++ b/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
@@ -64,7 +64,8 @@ class Input(CustomBaseModel):
     )
     input_type: InputType = Field(description="Type of input data (e.g., 'image').")
     image_layout: ImageLayout = Field(
-        description="Layout of the input image data (e.g., 'hwc').", default=None
+        description="Layout of the input image data (e.g., 'hwc').",
+        default=ImageLayout.CHW,
     )
     shape: List[int] = Field(
         min_length=1,
--- a/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
+++ b/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
@@ -1,8 +1,8 @@
 from typing import List, Optional
 
-from pydantic import Field
+from pydantic import Field, model_validator
 
-from ..enums import DataType, InputType
+from ..enums import DataType, InputType, InputImageLayout
 from .custom_base_model import CustomBaseModel
 
 
@@ -48,6 +48,9 @@ class Input(CustomBaseModel):
     @type input_type: InputType
     @ivar input_type: Type of input data (e.g., 'image').
 
+    @type input_image_layout: InputLayout
+    @ivar input_image_layout: Layout of the input image data (e.g., 'hwc').
+
     @type shape: list
     @ivar shape: Shape of the input data as a list of integers (e.g. [H,W], [H,W,C], [BS,H,W,C], ...).
 
@@ -60,6 +63,9 @@ class Input(CustomBaseModel):
         description="Data type of the input data (e.g., 'float32')."
     )
     input_type: InputType = Field(description="Type of input data (e.g., 'image').")
+    input_image_layout: InputImageLayout = Field(
+        description="Layout of the input image data (e.g., 'hwc').", default=None
+    )
     shape: List[int] = Field(
         min_length=1,
         max_length=5,
@@ -68,3 +74,17 @@ class Input(CustomBaseModel):
     preprocessing: PreprocessingBlock = Field(
         description="Preprocessing steps applied to the input data."
     )
+
+    @model_validator(mode="before")
+    def validate_anchors(
+        cls,
+        values,
+    ):
+        if (
+            values["input_type"] == InputType.IMAGE.value
+            and "input_image_layout" not in values.keys()
+        ):
+            raise ValueError(
+                "It's obligatory to specify input_image_layout for image input type."
+            )
+        return values
--- a/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
+++ b/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
@@ -83,26 +83,18 @@ class Input(CustomBaseModel):
     ):
         if "layout" in values.keys():
             values["layout"] = values["layout"].upper()
-            idx_c = values["layout"].find("C")
             idx_n = values["layout"].find("N")
 
             if len(values["layout"]) != len(values["shape"]):
                 raise ValueError("Layout and shape must have the same length.")
 
             if idx_n not in [-1, 0]:
-                raise ValueError(
-                    "N (batch size) must be the first letter if included."
-                )
+                raise ValueError("N (batch size) must be the first letter if included.")
 
             if values["input_type"] == InputType.IMAGE.value:
                 if "C" not in values["layout"]:
                     raise ValueError(
                         "C letter must be present in layout for image input type."
-                    )
-                
-                if values["shape"][idx_c] not in [1, 3]:
-                    raise ValueError(
-                        "Color dimension for image input type must either be 1 (grayscale) or 3 (color)."
                     )
 
             if len(values["layout"]) != len(set(values["layout"])):
--- a/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
+++ b/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
@@ -88,7 +88,9 @@ class Input(CustomBaseModel):
                 raise ValueError("Layout and shape must have the same length.")
 
             if values["layout"][0] != "N":
-                raise ValueError("First letter of layout must always be N (batch size).")
+                raise ValueError(
+                    "First letter of layout must always be N (batch size)."
+                )
 
             if values["input_type"] == InputType.IMAGE.value:
                 if "C" not in values["layout"]:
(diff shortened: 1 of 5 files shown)


luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -60,6 +63,9 @@ class Input(CustomBaseModel):
         description="Data type of the input data (e.g., 'float32')."
     )
     input_type: InputType = Field(description="Type of input data (e.g., 'image').")
+    image_layout: ImageLayout = Field(
```
kozlov721, line 66: Would it make sense to assume `"chw"` by default?

tersekmatija: We could go just with "layout".
I do think in most cases it will be an image input, so it would make sense doing a "BCHW" layout by default? I think it won't error out here, but it will fail downstream (e.g. at conversion) where we could potentially catch that and warn the user?

Not sure if we'd want to introduce other dependencies there, but we could partially validate this when generating the NN archive by reading ONNX or TFLITE input shapes?

luxonis_ml/nn_archive/config_building_blocks/enums/image_layout.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class ImageLayout(Enum):
+    """Represents the input image layout of the model."""
+
```
tersekmatija, line 6: Do we want to include batch dimension in the layout?

jkbmrz: I was thinking it might be easier for users to decide if we have both batched and non-batched options available (they might get lost if their models has 4 input dimensions while we only provide 3 here). But it's true that ignoring the batch dimension is more clean. Thus, I'm removing the batch dimension options in [d0652e9](https://github.com/luxonis/luxonis-ml/pull/109/commits/d0652e97619e97eaacd1432f729e71426587841e)

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
jkbmrz: > I think it won't error out here, but it will fail downstream (e.g. at conversion) where we could potentially catch that and warn the user?
That's true, I guess we don't have to worry about every possible mistake of the user (and rather warn them later if they did something wrong). I'm adding the default layout in [217a33f](https://github.com/luxonis/luxonis-ml/pull/109/commits/217a33f4cfefbe733ed0ca5ac3fd2bade2c56339)

jkbmrz: > we could partially validate this when generating the NN archive by reading ONNX or TFLITE input shapes?

How would you go about that? The input shape doesn't provide information on HWC/CHW.

jkbmrz: > We could go just with "layout".

I named it `image_layout` intentionally to make things less confusing when adding a model with `InputType.RAW` with input layout different from the HWC/CHW. If we rename it to "layout", would make sense to add additional values to the ImageLayout enum that would be more suitable for the `InputType.RAW` cases (e.g. "custom")?

tersekmatija: > How would you go about that? The input shape doesn't provide information on HWC/CHW.

I think mainly on ranks/dimensions and then we could have some checks which issue warnings if channel is not significantly larger or smaller than HW.

I understand intention behind `image_layout`, and based on this conversation we would have to add more options. Alternatively, we don't treat it as enum and rather a string. Or a dedicated object for which we validate certain configurations (e.g. order of C against H,W; multiple repetitions of H are not allowed; ...).

luxonis_ml/nn_archive/config_building_blocks/enums/image_layout.py
tersekmatija: Sorry, should have been more clear. I meant if we should enforce batch dimension on all inputs?

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
jkbmrz: > I think mainly on ranks/dimensions and then we could have some checks which issue warnings if channel is not significantly larger or smaller than HW.

If I understand correctly, we could simply add an additional model_validator to the Input class doing the checks on `image_layout` and `shape` (no need to validate this when generating the NN archive).

jkbmrz: > I understand intention behind image_layout, and based on this conversation we would have to add more options. Alternatively, we don't treat it as enum and rather a string. Or a dedicated object for which we validate certain configurations (e.g. order of C against H,W; multiple repetitions of H are not allowed; ...).

What more options do you have in mind? Also, I'm not sure I get the the dedicated object part.

luxonis_ml/nn_archive/config_building_blocks/enums/image_layout.py
jkbmrz: I see. If that's desirable, we can force it by an additional `model_validator` to the `Input` class doing the checks on `input_type` and `shape` fields - if `InputType.IMAGE`, then `shape` must be of length=4.

conorsim (commented):
Looks good generally but leaving a couple comments

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -68,3 +75,17 @@ class Input(CustomBaseModel):
     preprocessing: PreprocessingBlock = Field(
         description="Preprocessing steps applied to the input data."
     )
+
+    @model_validator(mode="before")
+    def validate_anchors(
```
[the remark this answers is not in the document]
jkbmrz: This is a mistake. Thanks for finding!

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
jkbmrz: I've switched to textual layout in [d5476aa](https://github.com/luxonis/luxonis-ml/pull/109/commits/d5476aa083eac8d602a1ad898f0f4ca731ecadad). For validation, I'd still propose to do it with model_validator inside the `Input` instead of at the time of nn archive creation. In my logic, user can now input something random as layout, e.g. "skoQSKO/?!QSqK", which makes the config invalid (OR is there some reason why we should allow this and only fail later?). I'd include the following tests:
- fail if len(layout) =! len(shape)
- fail if layer doesn't start with N
- fail if layer has letter repetition
- fail if dimension of C =! 1 or 3

luxonis_ml/nn_archive/config_building_blocks/enums/image_layout.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+
+
+class ImageLayout(Enum):
```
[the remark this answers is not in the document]
jkbmrz: The enum was removed in commit [d5476aa](https://github.com/luxonis/luxonis-ml/pull/109/commits/d5476aa083eac8d602a1ad898f0f4ca731ecadad) (switching to textual definition of layout)

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
tersekmatija: Sorry, I see what you mean. Yeah let's already validate there!

jkbmrz: Added in [2fd6907](https://github.com/luxonis/luxonis-ml/pull/109/commits/2fd69074e8e55c4f160d8f28b5a9b1042a93f3a5).

luxonis_ml/nn_archive/config_building_blocks/enums/image_layout.py
jkbmrz: Added in [2fd6907](https://github.com/luxonis/luxonis-ml/pull/109/commits/2fd69074e8e55c4f160d8f28b5a9b1042a93f3a5).

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -65,6 +68,42 @@ class Input(CustomBaseModel):
...
+        cls,
+        values,
+    ):
+        if "layout" in values.keys():
+            values["layout"] = values["layout"].upper()
+
+            if len(values["layout"]) != len(values["shape"]):
+                raise ValueError("Layout and shape must have the same length.")
+
+            if values["layout"][0] != "N":
```
tersekmatija, line 90: Not sure if we want to always enforce it, but might be good to just check that if it is present, it'd be in the first place?

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -65,6 +68,42 @@ class Input(CustomBaseModel):
...
+                    "First letter of layout must always be N (batch size)."
+                )
+
+            if values["input_type"] == InputType.IMAGE.value:
+                if "C" not in values["layout"]:
+                    raise ValueError(
+                        "C letter must be present in layout for image input type."
+                    )
+                idx_c = values["layout"].find("C")
+                if values["shape"][idx_c] not in [1, 3]:
```
tersekmatija, line 101: I think we should remove this (1,3) check as well as you might have different inputs also for downstream models that operate on features.

tersekmatija (changes requested):

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -65,6 +68,42 @@ class Input(CustomBaseModel):
         max_length=5,
         description="Shape of the input data as a list of integers (e.g. [H,W], [H,W,C], [BS,H,W,C], ...).",
     )
+    layout: str = Field(
+        description="Lettercode interpretation of the input layout (e.g., 'HCHW').",
```
kozlov721, line 72: Typo, should be `'NCHW'`

jkbmrz: Fixed in [fdd6bff](https://github.com/luxonis/luxonis-ml/pull/109/commits/fdd6bff9eee194d210a767e258c1e47cd2f25986). Thanks!

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
jkbmrz: Fixed in [fdd6bff](https://github.com/luxonis/luxonis-ml/pull/109/commits/fdd6bff9eee194d210a767e258c1e47cd2f25986). Thanks!

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
```diff
@@ -51,6 +51,9 @@ class Input(CustomBaseModel):
     @type shape: list
     @ivar shape: Shape of the input data as a list of integers (e.g. [H,W], [H,W,C], [BS,H,W,C], ...).
 
+    @type layout: str
+    @ivar layout: Lettercode interpretation of the input data dimensions (e.g., 'HCHW').
```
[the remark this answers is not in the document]
jkbmrz: Fixed in [a6d30c2](https://github.com/luxonis/luxonis-ml/pull/109/commits/a6d30c2be99ec8a0c169a510055fe3b33481685b). Thanks!

luxonis_ml/nn_archive/config_building_blocks/base_models/input.py
jkbmrz: Fixed in [c79c200](https://github.com/luxonis/luxonis-ml/pull/109/commits/c79c200fcdac9fa75d8aa969a94a64f368ccca15). Thanks!

tersekmatija (approved):

conorsim (approved):

PR #110, opened 2024-05-27
Annotations Refactor

kozlov721:
# Annotations Refactor

- Refactoring of how the annotations are handled and defined
- Changed implementation of label groups
- Support for images without annotations

### LDF Generator Output

- Annotations are a separate field in the output dictionary and are optional
  - This makes it possible to add only images without annotations to the LDF
- The annotation field is more structured instead of just a "value" field.
- Added support for an `"instance_id"` field in the annotations for instance-aware task
  - For example; with keypoint detection, the bounding boxes and keypoint labels must be ordered the same way. This was not guaranteed before, the individual instances were loaded in the order they were added to the dataset.  
  
**Before**
```python
yield {
    "file": img_path,
    "class": "dog",
    "type": "box",
    "value": (x, y, w, h)
}
```

**After**
```python
yield {
    "file": img_path,
    "annotation": {
        "type": "boundingbox",
        "class": "dog",
        "instance_id": 1,
        "x": x,
        "y": y,
        "w": w,
        "h": h,
    },
}
```

### Structure of the label dictionary

The label dictionary is now a mapping from task names to tuples of `numpy` arrays and label types instead of a nested dictionary of labels for each label type.

By default, the tasks are just a string representation of the `LabelType`, but they can be arbitrary. This makes possible to have more than just one label of a particular `LabelType` in the dictionary.

**Before**
```python
labels: dict[str, dict[LabelType, np.ndarray]] = {
    "default": { # default task group
        LabelType.CLASSIFICATION: class_array,
        LabelType.BOUNDINGBOX: bbox_array,
        LabelType.SEGMENTATION: seg_array,
    }
}
```

**Now**
```python
labels: dict[str, tuple[np.ndarray, LabelType]] = {
    "classification": (class_array, LabelType.CLASSIFICATION),
    "boundingbox": (bbox_array, LabelType.CLASSIFICATION),
    "segmentation": (seg_array, LabelType.SEGMENTATION),
}
```

**Hypothetical use case for tasks**

Model with 1 detection and 2 classification heads would make use of the following label dictionary:
```python
{
    "color": (..., LabelType.CLASSIFICATION),
    "shape": (..., LabelType.CLASSIFICATION),
    "position": (..., LabelType.BOUNDINGBOX), 
}
```

### Miscellaneous

- `LabelType.KEYPOINT` renamed to `LabelType.KEYPOINTS`
- `"instance_id"` in parquet file renamed to `"uuid"` as to not collide with the new `"instance_id"` annotation attribute
- It is not necessary to call `set_classes`, the classes are inferred from the annotations
- Setting skeletons for keypoints is no longer mandatory
- Removed `ValAugmentations` and `TrainAugmentations` classes. Instead, the base `Augmentations` takes `only_normalize` argument which can be set to `True` for val augmentations 
- Added option to pass a `yaml` file with augmentations to `luxonis_ml data inspect` command


luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -15,14 +15,14 @@
 import rich.progress
 
 import luxonis_ml.data.utils.data_utils as data_utils
-from luxonis_ml.enums import AnnotationType
 from luxonis_ml.utils import LuxonisFileSystem, environ
 
-from ..utils.constants import LABEL_TYPES, LDF_VERSION
-from ..utils.enums import BucketStorage, BucketType, ImageType, MediaType
+from ..utils.constants import LDF_VERSION
```
conorsim, line 20: Should we bump `LDF_VERSION`? Might miss it if we did going through, but I don't see changes to utils.constants

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -54,28 +54,18 @@ def __init__(
             C{dataset_name} nor C{dataset_id} are provided.
         """
 
-        if dataset_name is not None and not isinstance(dataset_name, str):
-            raise ValueError("`dataset_name` argument must be a string")
-
-        if dataset_name is None and dataset_id is None:
-            raise ValueError(
-                "Must provide either dataset_name or dataset_id when initializing LuxonisDataset"
-            )
```
conorsim, line 63: Is this validation handled elsewhere?

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -411,37 +302,40 @@ def get_classes(
...
         return classes, classes_by_task
 
-    def get_skeletons(self, sync_mode: bool = False) -> Dict[str, Dict]:
-        """Returns the dictionary defining the semantic skeleton for each class using
-        keypoints.
+    def set_skeletons(
+        self, skeletons: Dict[str, Dict], task: Optional[str] = None
+    ) -> None:
+        if task is None:
+            raise NotImplementedError("Skeletons must be set for a specific task")
```
conorsim, line 315: Should we just make `task` a required argument then?

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -303,105 +262,37 @@ def update_source(self, source: LuxonisSource) -> None:
...
-            if task not in LABEL_TYPES:
-                raise Exception(f"Task {task} is not a supported task")
             self.datasets[self.dataset_name]["classes"][task] = classes
         else:
-            for task in LABEL_TYPES:
-                self.datasets[self.dataset_name]["classes"][task] = classes
+            raise NotImplementedError(
+                "Setting classes for all tasks not yet supported. "
+                "Set classes individually for each task"
+            )
```
conorsim, line 276: Should we change `task` to be required instead?

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -529,76 +370,116 @@ def _add_process_batch(batch_data: List[Annotation]) -> None:
...
+            if i % batch_size == 0:
                 _add_process_batch(batch_data)
                 batch_data = []
 
         _add_process_batch(batch_data)
 
+        _, curr_classes = self.get_classes()
+        if not curr_classes:
+            for task, classes in classes_per_task.items():
+                self.set_classes(list(classes), task)
```
conorsim, line 474: Would this code work automatically if we call `add` many times on the same dataset? Or would it only detect changes for the first time?

luxonis_ml/data/parsers/luxonis_parser.py
```diff
@@ -114,16 +114,16 @@ def __init__(
         self.dataset = self.dataset_constructor(dataset_name=dataset_name, **kwargs)
         self.parser = self.parsers[self.dataset_type](self.dataset)
 
-    def parse(self, **kwargs) -> BaseDataset:
```
conorsim, line 117: I changed this from `LuxonisDataset` -> `BaseDataset` previously since if we use a custom `dataset_plugin` we could have another class than `LuxonisDataset`

conorsim (commented):
This looks like an amazing refactor!

Commenting on some mostly minor things for now, but here are some bigger notes
- Should we bump the major version of `LDF_VERSION` since it's not backwards compatible?
- Would the lack of backwards compatibility cause any issues for others?
- When I try running the parser example notebook, I get `KeyError: 'boundingbox'` and when I `luxonis_ml data inspect coco_test` I see there are no annotations. Just noting in case there could be a bug with parsers

kozlov721:
>Should we bump the major version of LDF_VERSION since it's not backwards compatible?

Yes, I'll update the version.

>Would the lack of backwards compatibility cause any issues for others?

I want to merge this PR only after a PR with full support in `luxonis-train` is also ready to merge and then immediately release a new version of both `luxonis-ml` and `luxonis-train`, so we can just refer people to the latest versions to solve all compatibility issues. 

>When I try running the parser example notebook, I get KeyError: 'boundingbox' and when I luxonis_ml data inspect coco_test I see there are no annotations. Just noting in case there could be a bug with parsers

This is strange, the notebooks work fine on my end. Maybe try to delete the downloaded coco dataset you have locally in case it is corrupted somehow.

luxonis_ml/data/parsers/luxonis_parser.py
kozlov721: Gotcha, that makes sense

luxonis_ml/data/datasets/luxonis_dataset.py
kozlov721: No, this fell through the cracks. I'll add it back.

luxonis_ml/data/datasets/annotation.py
```diff
@@ -0,0 +1,382 @@
...
+    @staticmethod
+    def combine_to_numpy(
+        annotations: List["RLESegmentationAnnotation"],
+        class_mapping: Dict[str, int],
+        height: int,
+        width: int,
+    ) -> np.ndarray:
+        seg = np.zeros((len(class_mapping), height, width))
+        for ann in annotations:
+            assert isinstance(ann.counts, bytes)
```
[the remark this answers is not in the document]
kozlov721: No, after loading the annotation from parquet it will always be bytes. The `assert` is just to please the type checker.

luxonis_ml/data/datasets/luxonis_dataset.py
kozlov721: Same as above

luxonis_ml/data/parsers/luxonis_parser.py
kozlov721: In that case we'd need to add `make_splits` to the `BaseDataset` interface.

luxonis_ml/data/datasets/luxonis_dataset.py
kozlov721: Yeah, it would only detect the classes for the first time. So if we'd ingest an additional class, the classes would have to be set manually. I can improve it so it detects a new class and adds it automatically too.

conorsim: > I can improve it so it detects a new class and adds it automatically too.

It would be nice-to-have but not strictly necessary for the PR IMO

luxonis_ml/data/parsers/luxonis_parser.py
conorsim: > In that case we'd need to add make_splits to the BaseDataset interface.

I think we can do that. It was probably an oversight that it wasn't already there

luxonis_ml/data/datasets/annotation.py
```diff
@@ -0,0 +1,385 @@
...
+        pass
+
+
+class ClassificationAnnotation(Annotation):
+    _label_type = LabelType.CLASSIFICATION
+    type_: Literal["classification"] = Field("classification", alias="type")
+
+    @staticmethod
+    def combine_to_numpy(
+        annotations: List["BBoxAnnotation"], class_mapping: Dict[str, int], **_
```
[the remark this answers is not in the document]
kozlov721: Yes, thanks!

jkbmrz (approved):

luxonis_ml/data/__main__.py
```diff
@@ -131,38 +139,74 @@ def inspect(
...
+
+    augmentations = None
+    if aug_config is not None:
+        with open(aug_config) as file:
+            config = (
+                yaml.safe_load(file)
+                if aug_config.suffix == ".yaml"
+                else json.load(file)
+            )
+        augmentations = Augmentations([512, 512], config)
```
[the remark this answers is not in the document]
kozlov721: I can add it to arguments, but this command is meant to be used just for a quick debugging so it shouldn't matter much.

tersekmatija:
> so we can just refer people to the latest versions to solve all compatibility issues.

Can we have on dev branches to test to make sure everything works as expected?

luxonis_ml/data/datasets/base_dataset.py
```diff
@@ -205,32 +105,43 @@ def get_skeletons(self) -> Dict[str, Dict]:
         pass
 
     @abstractmethod
-    def add(self, generator: DatasetIterator) -> None:
+    def add(self, generator: DatasetIterator, batch_size: int = 1_000_000) -> None:
```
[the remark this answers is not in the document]
kozlov721: I don't know why it's set to 1_000_000 by default (CC @conorsim). It's just an internal thing for how often for we write to the parquet files.

luxonis_ml/data/augmentations/utils.py
```diff
@@ -26,6 +33,14 @@ def __init__(self, train_rgb: bool = True):
         self.is_batched = False
         self.aug_batch_size = 1
 
+        self.batch_transform, self.spatial_transform = self._parse_cfg(
+            image_size=image_size,
+            augmentations=[a for a in augmentations if a["name"] == "Normalize"]
```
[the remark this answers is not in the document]
kozlov721: Then no augmentations are performed.

luxonis_ml/data/datasets/annotation.py
```diff
@@ -0,0 +1,385 @@
...
+    @type instance_id: int
+    @ivar instance_id: The instance id of the annotation. This determines the order in
+        which individual instances are loaded in L{LuxonisLoader}.
+    @type _label_type: ClassVar[L{LabelType}]
+    @ivar _label_type: The label type of the annotation.
+    """
+
+    _label_type: ClassVar[LabelType]
+
+    task: str
```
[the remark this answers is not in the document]
kozlov721: The label dictionary is now indexed by the `"task"` field instead of by the `LabelType`. This allows to have more annotations of the same type in the labels.

luxonis_ml/data/datasets/base_dataset.py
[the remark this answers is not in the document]
conorsim: Yeah I believe the goal is to avoid OOM for really large ingestion. The choice of batch size is a bit arbitrary

luxonis_ml/data/datasets/annotation.py
```diff
@@ -0,0 +1,385 @@
...
+    ) -> np.ndarray:
+        keypoints = np.zeros((len(annotations), len(annotations[0].keypoints) * 3 + 1))
+        for i, ann in enumerate(annotations):
+            keypoints[i] = ann.to_numpy(class_mapping)
+        return keypoints
+
+
+class SegmentationAnnotation(Annotation):
+    """Base class for segmentation annotations."""
+
```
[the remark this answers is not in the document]
kozlov721: This is just a parent class, `type_` is specified in its children.

PR #112, opened 2024-06-03
Single head class for YOLO models

kkeroo:
Head classes for `HeadInstanceSegmentationYOLO`, `HeadKeypointDetectionYOLO`, `HeadOBBDetectionYOLO`, and `HeadObjectDetectionYOLO` are grouped into one class `HeadObjectDetectionYOLO`. The class includes optional parameters for different tasks.

To validate the model we specify which parameters are required and unsupported in a dictionary and then check for specific family if all parameters are present. Additionally, output validation is added to validate if defined `outputs` is appropriate for a specific family.


kkeroo:
Tests are added for classification head, all YOLO heads, and SSD head. Tests for ObjectDetectionHead and SegmentationHead are not added because they are included in other heads. Commit [67b1444](https://github.com/luxonis/luxonis-ml/pull/112/commits/67b1444895f1b7d25881743ce628213fce49a884)

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
```diff
@@ -190,93 +148,146 @@ def validate_label_type(
...
+    @ivar n_keypoints: Number of keypoints per bbox in YOLO keypoint detection.
+    @type is_softmax: bool | None
+    @ivar is_softmax: True, if output is already softmaxed in YOLO instance segmentation.
     """
 
-    family: Literal["InstanceSegmentationYOLO"] = Field(
-        ..., description="Decoding family."
-    )
-    outputs: OutputsInstanceSegmentationYOLO = Field(
+    family: Literal[
```
tersekmatija, line 170: Do those to actually need to be explicitly defined? I think we limit ourselves to only one type of output, whereas we could have family be just "YOLO" or "DetectionYOLO" and output some general OutputsYOLO which can have additional fields on the output?

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
```diff
@@ -190,93 +148,146 @@ def validate_label_type(
...
+        required_fields = {
+            "InstanceSegmentationYOLO": [
+                "postprocessor_path",
+                "n_prototypes",
+                "is_softmax",
+            ],
+            "KeypointDetectionYOLO": ["n_keypoints"],
+        }
+
+        unsupported_fields = {
```
tersekmatija, line 232: This invalidates instance segmentation with keypoints. We don't support training of this for now, so it's fine, just something to note.

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
```diff
@@ -190,93 +148,146 @@ def validate_label_type(
...
-    family: Literal["OBBDetectionYOLO"] = Field(..., description="Decoding family.")
-    outputs: OutputsOBBDetectionYOLO = Field(
-        description="A configuration specifying which output names from the `outputs` block of the archive are fed into the head."
-    )
+    @model_validator(mode="before")
+    def validate_anchors(cls, values):
+        if (
+            "anchors" in values
+            and values["anchors"] is not None
+            and values["subtype"] == ObjectDetectionSubtypeYOLO.YOLOv6
```
tersekmatija, line 268: I believe v8 as well?

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
```diff
@@ -190,93 +148,146 @@ def validate_label_type(
...
 
-    @field_validator("family")
-    def validate_label_type(
-        cls,
-        value,
-    ):
-        if value != "OBBDetectionYOLO":
-            raise ValueError("Invalid family")
-        return value
+    @model_validator(mode="before")
```
tersekmatija, line 273: This can perhaps be omitted if we can make outputs generic.

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
kkeroo: Added in [e6ddc8e](https://github.com/luxonis/luxonis-ml/pull/112/commits/e6ddc8ec4d9199bb47d82535b71fe6fba3c3b6cf).

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
kkeroo: Okay, I made a generic YOLO head with optional parameters and a generic Yolo output with optional parameters. Validation is performed based on the defined attributes and then the Output attributes are also validated to match the specific task. This introduces more validation complexity. Added in [e6ddc8e](https://github.com/luxonis/luxonis-ml/pull/112/commits/e6ddc8ec4d9199bb47d82535b71fe6fba3c3b6cf).

luxonis_ml/nn_archive/config_building_blocks/base_models/head.py
kkeroo: One generic output class [e6ddc8e](https://github.com/luxonis/luxonis-ml/pull/112/commits/e6ddc8ec4d9199bb47d82535b71fe6fba3c3b6cf).

tersekmatija (approved):
I think it looks good. I see currently we still only select one specific task on the Yolo head. I think this is fine for the time being. But with generic detection head in luxonis-train in the future, we will have to somehow adapt this to support multiple tasks. CC @kozlov721 if any extra thoughts there.

conorsim:
It looks good, thanks for adding tests!

> I see currently we still only select one specific task on the Yolo head. I think this is fine for the time being.

I think we maybe want a plan for this though. How would we support instance segmentation with keypoints in the future? Would we add additional validation logic for `instance_segmentation_keypoints`? Or is there some way we can better generalize different combinations?

tersekmatija:
I believe one option would be to just not define different ones - you should be able to infer additional tasks based on the outputs/other fields?

kkeroo:
#### Support for multiple tasks
We can now define more than one task in YOLO head. Added in [b8add61](https://github.com/luxonis/luxonis-ml/pull/112/commits/b8add61a9aadfa36e12b5f5c8d1990da2ca7832d).

The validator checks if all the required fields for specific tasks are present in defined fields. Additionally, to avoid any unnecessary fields, the validator also checks if any defined fields are not supported by the inferred tasks. So, this way we cannot have any extra defined fields.

The `OutputsYOLO` is validated similarly.

One additional test is added for instance segmentation with keypoints.

tersekmatija (approved):

PR #113, opened 2024-06-10
add OCR Decoding support - WIP

N950:
This is a WIP to add CTC OCR recognition/decoding
Conformity to contribution guidelines will be fixed before closing

Next change will be adding kpt label to the TEXT LabelType, even though the goal is only OCR recognition, on the data side it makes since to create the annotation/LabelType from the begining to support kpt annotations

### DIFF
--- a/luxonis_ml/data/augmentations/custom/ocr.py
+++ b/luxonis_ml/data/augmentations/custom/ocr.py
@@ -1,0 +1,121 @@
+from typing import Tuple
+import random
+import albumentations as A
+
+from albumentations.core.transforms_interface import ImageOnlyTransform
+
+import numpy as np
+from ..utils import AUGMENTATIONS
+
+
+@AUGMENTATIONS.register_module()
+class OCRAugmentation(ImageOnlyTransform):
+
+    def __init__(self, image_size: Tuple[int, int], is_rgb: bool, is_train: bool):
+        """
+
+        @param image_size: OCR model input shape.
+        @type image_size: Tuple[int, int]
+        @param is_rgb: True if image is RGB. False if image is GRAY.
+        @type is_rgb: bool
+        @param is_train: True if image is train. False if image is val/test.
+        @type is_train: bool
+        """
+        super(OCRAugmentation, self).__init__()
+        self.transforms = A.Compose(
+            transforms=[
+                A.OneOf(
+                    transforms=[
+                        A.RandomScale(
+                            scale_limit=(-0.5, 0),  # (1 + low, 1 + high) => (0.5, 1) : random downscale (min half size)
+                            always_apply=True,
+                            p=1
+                        ),
+                        A.MotionBlur(
+                            blur_limit=(19, 21),
+                            p=1.0,
+                            always_apply=True,
+                            allow_shifted=False
+                        ),
+                        A.Affine(
+                            translate_percent=(0.05, 0.07),
+                            scale=(0.7, 1.0),
+                            rotate=(-7, 7),
+                            shear=(5, 35),
+                            always_apply=True,
+                            p=1
+                        )
+                    ],
+                    p=0.2
+                ),
+                A.OneOf(
+                    transforms=[
+                        A.ISONoise(
+                            color_shift=(0.01, 0.1),
+                            intensity=(0.1, 1.0),
+                            always_apply=True
+                        ),
+                        A.GaussianBlur(
+                            blur_limit=(7, 9),  # kernel
+                            sigma_limit=(0.1, 0.5),
+                            always_apply=True,
+                            p=0.2
+                        ),
+                        A.ColorJitter(
+                            brightness=(0.11, 1.0),
+                            contrast=0.5,
+                            saturation=0.5
+                        )
+                    ],
+                    p=0.2
+                ),
+                A.Compose(  # resize to image_size with aspect ratio, pad if needed
+                    transforms=[
+                        A.LongestMaxSize(max_size=max(image_size), interpolation=1),
+                        A.PadIfNeeded(min_height=image_size[0], min_width=image_size[1], border_mode=0,
+                                      value=(0, 0, 0))
+                    ]
+                ),
+                A.Normalize(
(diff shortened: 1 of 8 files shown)


luxonis_ml/data/datasets/annotation.py
```diff
@@ -92,6 +93,39 @@ def combine_to_numpy(
...
+            text_label[idx] = class_
+        return text_label
+
+    @staticmethod
+    def combine_to_numpy(
+        annotations: List["TextAnnotation"],
+        class_mapping: Dict[str, int],
+        **_,
+    ) -> np.ndarray:
+        text_labels = None
```
kozlov721, line 120: `text_labels` can be already set to `np.zeros((len(annotations), ann.max_len))` so there's no chance to return `None`

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -174,6 +174,7 @@ def _load_image_with_annotations(self, idx: int) -> Tuple[np.ndarray, Labels]:
 
         uuid = self.instances[idx]
         df = self.df.loc[uuid]
+        print(df)
```
kozlov721, line 177: forgotten `print`

luxonis_ml/data/utils/data_utils.py
```diff
@@ -27,3 +28,44 @@ def _check_valid_array(path: str) -> bool:
...
+
+    @param classes: valid character classes.
+    @type classes:
+    @type value: str
+    @param value: text value to validate.
+    @rtype: str
+    @return: same input value if it's valid, raises a Warning and cleans/ignores invalid characters.
+    """
+    clean_value = ""
+    text_value, max_len = value
```
kozlov721, line 47: Is it guaranteed the `value` will always contain exactly 2 characters? Even if so, maybe going with `Tuple[str, str]` would be better. In the very least documenting the 2-char requirement.

luxonis_ml/data/utils/data_utils.py
```diff
@@ -27,3 +28,44 @@ def _check_valid_array(path: str) -> bool:
             )
         if not _check_valid_array(value):
             raise Exception(f"Array at path {value} is not a valid numpy array (.npy)")
+
+
+def validate_text_value(
+    value: str,
```
kozlov721, line 34: The annotation of `value` seems to be incorrect.

kozlov721 (commented):
Left some comments, otherwise looks good.

luxonis_ml/data/augmentations/custom/ocr.py
```diff
@@ -0,0 +1,121 @@
...
+
+        @param image_size: OCR model input shape.
+        @type image_size: Tuple[int, int]
+        @param is_rgb: True if image is RGB. False if image is GRAY.
+        @type is_rgb: bool
+        @param is_train: True if image is train. False if image is val/test.
+        @type is_train: bool
+        """
+        super(OCRAugmentation, self).__init__()
+        self.transforms = A.Compose(
```
klemen1999, line 25: Is this a set of some standard augmentations that are usually performed for OCR task or how is this defined?

luxonis_ml/data/augmentations/custom/ocr.py
```diff
@@ -0,0 +1,121 @@
...
+                        ),
+                        A.ColorJitter(
+                            brightness=(0.11, 1.0),
+                            contrast=0.5,
+                            saturation=0.5
+                        )
+                    ],
+                    p=0.2
+                ),
+                A.Compose(  # resize to image_size with aspect ratio, pad if needed
```
klemen1999, line 72: Resize and Normalize are already part of the default augmentations. Resize is always [done](https://github.com/luxonis/luxonis-ml/blob/7981992f476f4ba5ef4f63c4329415fc40f8b00a/luxonis_ml/data/augmentations/utils.py#L66-L72) (you can control if it keeps aspect ratio or not) and Normalize is also appended to list of augmentations (if used by luxonis-train, can be deactivated through config though). So is this needed here?

luxonis_ml/data/augmentations/custom/ocr.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+
+        @param image_size: OCR model input shape.
+        @type image_size: Tuple[int, int]
+        @param is_rgb: True if image is RGB. False if image is GRAY.
+        @type is_rgb: bool
+        @param is_train: True if image is train. False if image is val/test.
+        @type is_train: bool
+        """
+        super(OCRAugmentation, self).__init__()
```
kozlov721, line 24: Le'ts keep it just `super().__init__()`. The arguments in `super` are a relic from python 2.

luxonis_ml/data/augmentations/custom/ocr.py
conorsim: Also curious on this

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -292,6 +293,9 @@ def set_classes(self, classes: List[str], task: Optional[str] = None) -> None:
             self.fs.put_file(local_file, "metadata/classes.json")
             self._remove_temp_dir()
 
+    def set_global_metadata(self, metadata: Dict[str, Any]) -> None:
+        self.global_metadata = metadata
```
conorsim, line 297: Due to GCS datasets, I think we need a way to persist this via storage instead of just memory? Perhaps we could use the existing `datasets.json` or `metadata` folder?

conorsim (commented):
> Next change will be adding kpt label to the TEXT LabelType, even though the goal is only OCR recognition, on the data side it makes since to create the annotation/LabelType from the begining to support kpt annotations

What is meant by this? We have the `LabelType.KEYPOINTS` already. We also plan to support nested annotations, so I think the final form for OCR + keypoints would be TEXT and KEYPOINTS nested within a BOUNDINGBOX

klemen1999:
Closed as it is out of date and now supported otherwise

PR #116, opened 2024-06-17
COCO Parser Speedup

JSabadin:
Modified the code for building the COCO dataset by using dictionaries for quick lookups, rather than repeatedly scanning the entire list of images and annotations.

Dictionaries provide an average-case constant time complexity (O(1)) for lookups, in contrast to the linear time complexity (O(n)) of list scans. This optimization considerably reduces the overall time complexity of the code.

The time required to create the **_COCO train dataset_**—prior to sending the batch to the bucket—decreased **_from 2 hours to 20 minutes_**.


kozlov721 (approved):
LGTM

conorsim (approved):
LGTM

PR #125, opened 2024-07-02
Nested Annotations Support

JSabadin:
# Nested Annotations Support

It can work with a generator yielding like this:

```
yield {
    "file": path,
    "annotation": {
        "type": "detection",
        "class": "person",
        "bounding_box": {
            "type": "boundingbox",
            "x": x,
            "y": y,
            "w": w,
            "h": h,
        },
        "keypoints": {
            "type": "keypoints",
            "keypoints": keypoints
        }
    }
}
```

I am very open to discussing how this can be made better or more efficient, as I was relatively new to luxonis-ml when I started this task. @kozlov721, what improvements can be made so I can address them in the next commits?

If someone prefers to create a dataset without nested annotations, it is possible.


kozlov721:
I think it can be done much simpler if we approach the nested annotations just as a syntax sugar.
Then we'd only need to implement some internal wrapper around the iterable passed to `LuxonisDataset.add` that would unpack the nested annotations. 

I think it could look a bit like this:

**`luxonis_dataset.py`**
```python
def _add_generator_wrapper(generator):
    def create_new_record(record, annotation)
        ...

    for record in generator:
        ann = record["annotation"]
        if ann["type"] != "detection":
            yield record
        else:
            bbox = None
            for sub_ann_key in ["boundingbox", "segmentation", "keypoints"]:
                sub_ann = ann[sub_ann_key]
                if sub_ann_key != "boundinbox" and ann["scaled_to_boxes"]:
                    sub_ann = rescale_values(bbox, record["file"], sub_ann)
                elif sub_ann_key == "boundingbox":
                    bbox = sub_ann 
                new_ann = {
                    "instance_id": ann["instance_id"],
                    "class": ann["class_]",
                     "task":  sub_ann.pop("task", f"{ann['task']}-{sub_ann_key}"),
                     **sub_ann,
                }
                yield create_new_record(record, new_ann)
```

**`LuxonisDataset.add`**
```python
def add(self, generator, ...):
    generator = _add_generator_wrapper(generator)
    ...
```

Nothing more should be needed to change or add in this case, I think. 

A record with the detection annotation in the generator could look like:
```python
yield {
    "file": filepath.png,
    "annotation": {
        "instance_id": 1,
        "task": animals,
        "class": "dog",
        "scaled_to_boxes": True,
        "bbox": {"x": 0.5, "y": ...},
        "keypoints": [(0.1, 0.15, 2), ...],
        "segmentation": {
            "type": "polylines",
            "points": [...],
        }
    }
}
```
This would just internally get translated into 3 separate yields.

kozlov721 (changes requested):
I'd simplify the unpacking of nested annotations, so it happens already at the generator level. Otherwise looks good

JSabadin:
Thank you for the great feedback @kozlov721. Also, should the normalization step be performed in the generator? We can normalize the keypoints with bounding boxes, but we need the information about the width and height of the image to normalize the bounding boxes.

kozlov721:
> Thank you for the great feedback @kozlov721. Also, should the normalization step be performed in the generator? We can normalize the keypoints with bounding boxes, but we need the information about the width and height of the image to normalize the bounding boxes.

Yeah you will need the width and height of the image, but you do know the file name at this point (`record["file"]`). Then if you use `PIL`, you can read the size efficiently.

JSabadin:
Many annotations have keypoints outside of the bounding boxes. In such cases, is it okay to just ignore the ones outside the bounds of the bounding box? Otherwise, it will produce annotations outside the interval [0,1] when rescaling.

kozlov721 (commented):
Looks good. Could you also add some test cases for this?

kozlov721 (approved):
LGTM

conorsim (approved):
LGTM

PR #136, opened 2024-07-23
NN Archive - More Data Types

kozlov721:
Added more datatypes to be fully compatible with `modelconverter`.


klemen1999:
Addressed in #200

PR #137, opened 2024-07-25
Enhanced COCO Parser

klemen1999:
Added `COCOImgWithAnn` parser that can be used for parsing datasets in COCO format exported from Roboflow (file structure is a bit different than COCO exported from Fiftyone).
The thinking behind the name is that unlike Fiftyone format which has "data" directory with images and labels.json for each split, this one has annotations and images all in one directory. Though if you have a better idea for a name I'm happy to change it.


luxonis_ml/data/parsers/coco_parser.py
```diff
@@ -216,6 +217,10 @@ def generator() -> DatasetIterator:
 
                     if "keypoints" in ann.keys():
                         kpts = np.array(ann["keypoints"]).reshape(-1, 3)
+                        # clip values inplace
+                        np.clip(kpts[:, 0], 0, img_w, out=kpts[:, 0])
+                        np.clip(kpts[:, 1], 0, img_h, out=kpts[:, 1])
```
conorsim, line 222: Do we always want to clip keypoints?

luxonis_ml/data/parsers/coco_img_with_ann_parser.py
```diff
@@ -0,0 +1,70 @@
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+
+from .base_parser import BaseParser
+from .coco_parser import COCOParser
+
+
+class COCOImgWithAnnParser(COCOParser):
+    """Parses directory with COCO annotations to LDF. Unlike COCOParser here images and
```
conorsim, line 9:
```suggestion
    """Parses directory with COCO annotations to LDF. Unlike COCOParser, images and
```

Little nitpick for read-ability

conorsim (approved):
I can't think of a great name either. One that comes to mind is `COCOSplitAnnParser`, since the annotations file is split into each dataset split.

kozlov721 (approved):

luxonis_ml/data/parsers/coco_parser.py
klemen1999: Yes, otherwise KeypointAnnotation fails since it expects [tuple of normalized floats](https://github.com/luxonis/luxonis-ml/blob/f3a94195a73a012e4c18d021dc9417b187ffd794/luxonis_ml/data/datasets/annotation.py#L165).

kozlov721: Hm, maybe it would make more sense to allow the keypoints to be in [-1, 1] range instead, but I'm not entirely sure what's the best approach here.

klemen1999:
> I can't think of a great name either. One that comes to mind is `COCOSplitAnnParser`, since the annotations file is split into each dataset split.

Hm yeah but Fiftyone (default) COCO also has annotations split for each datasets,. The only difference is that images are in a separate `data` dir inside split dir where as here images in same dir as label json

kozlov721:
>Though if you have a better idea for a name I'm happy to change it.

What about `COCORoboflow` and `COCOFiftyone`? CC @conorsim

klemen1999:
Now `COCOParser` can handle both formats. CC: @conorsim @kozlov721

conorsim (approved):
LGTM

PR #140, opened 2024-07-31
Annotation Loading Speed-Up

JSabadin:
# Optimized Image and Annotations Loading Function

## Overview

We optimized the `_load_image_with_annotations` function, which previously took 25 minutes per epoch on the entire COCO training dataset for the BBox task. After the optimization, we tested the performance of lading the dataset on the COCO dataset with Segmentations and Bounding Boxes. The results showed a significant improvement, with the unoptimized function taking 1151 seconds and the optimized function taking only 502 seconds, achieving a 2x faster performance.

## Performance Comparison

| Function Version | Time (seconds) |
| ---------------- | -------------- |
| Unoptimized      | 1151           |
| Optimized        | 502            |

## Additional notes

For detailed testing results, please refer to the following link: [Testing Results](https://app.clickup.com/t/86bzmkk83).


kozlov721 (approved):

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -161,11 +174,12 @@ def _load_image_with_annotations(self, idx: int) -> Tuple[np.ndarray, Labels]:
             present annotations
         """
 
-        uuid = self.instances[idx]
-        df = self.df.filter(pl.col("uuid") == uuid)
+        ann_indices = self.idx_to_df_row[idx]
+        ann_rows = [self.df.row(row) for row in ann_indices]
         if not self.dataset.is_remote:
-            img_path = list(df.select("original_filepath"))[0][0]
+            img_path = ann_rows[0][8]
```
conorsim, line 180: Minor comment, but I believe this is a tuple and it might be easier to read if we use some other data type like a dict, so we have some keys and more well-defined structure in the `ann_rows`. I'm not sure how likely it is, but in the case that the tuple element order changes, it would break everything.

luxonis_ml/data/requirements.txt
```diff
@@ -1,12 +1,12 @@
 numpy>=1.22
 tqdm>=4.64.1
-opencv-python==4.7.0.68
-opencv-python-headless==4.7.0.68
+opencv-python>=4.10.0
+opencv-python-headless>=4.10.0
```
conorsim, line 4: Out of curiosity, why did we increase this version?

conorsim (approved):
Left some minor comments but looks good.

luxonis_ml/data/requirements.txt
kozlov721: There is a significant improvement in `imread` speed. Version `4.10.0` is ~2 times faster than `4.7.0`

luxonis_ml/data/loaders/luxonis_loader.py
kozlov721: There is an option to add `named=True` to `self.df.row()`, which produces a dictionary instead of a tuple, but at some performance cost. IMHO having it as a tuple is okay, if we'd change the order of elements in the parquet file or anything, it will get caught in tests.

JSabadin:
This PR also fixes the Polars bug related to the filtering issue described [here](https://github.com/pola-rs/polars/issues/17526).

YOLOv8 takes 21 minutes with a batch size of 16, mosaic, and 4 workers on the entire COCO training dataset. Our code takes 18 minutes. YOLOv6 takes 21minutes. This was tested with the commit that fixes the order in which augmentations are applied ([here](https://github.com/luxonis/luxonis-ml/commit/44404bddf1f51eb76252277bdb378d9e178cec99)), as well as the commit that ignores the masks when they are not necessary ([here](https://github.com/luxonis/luxonis-ml/commit/5b80ec253476084468e0a855056a79a4f6b1f4e1)).

ritchie46:
> This PR also fixes the Polars bug related to the filtering issue described https://github.com/pola-rs/polars/issues/17526.

I think it is important to stress it isn't a Polars bug. It's a bug to fork a process that does any multi-threading. Spawn should be used.

PR #141, opened 2024-08-01
Custom Tasks in Parsers

kozlov721:
Added an option to add `task_mapping` attribute to the `LuxonisParser` for specifying custom task names.

**Example**:

```python
parser = LuxonisParser(..., task_mapping={LabelType.BOUNDINGBOX: "object_detection"})
dataset = parser.parse()
tasks = dataset.get_tasks()
print(dataset.get_tasks())
# >> ["object_detection", "classification"]
```

**CLI**:
```bash
luxonis_ml data parse path/to/data --task-name boundingbox=detection -tn segmentation=seg
```

### DIFF
--- a/luxonis_ml/data/parsers/luxonis_parser.py
+++ b/luxonis_ml/data/parsers/luxonis_parser.py
@@ -4,11 +4,8 @@ from enum import Enum
 from pathlib import Path
 from typing import Dict, Literal, Optional, Tuple, Type, Union
 
-from luxonis_ml.data import (
-    DATASETS_REGISTRY,
-    BaseDataset,
-    LuxonisDataset,
-)
+from luxonis_ml.data import DATASETS_REGISTRY, BaseDataset, LuxonisDataset
+from luxonis_ml.data.utils.enums import LabelType
 from luxonis_ml.enums import DatasetType
 from luxonis_ml.utils import LuxonisFileSystem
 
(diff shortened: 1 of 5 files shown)


kozlov721:
> @kozlov721 what would we do in a situation where we have multiple tasks of the same task type? For example, `dog-breed-classification` and `dog-color-classification`?

That's not a concern with parsers, we never have multiple takes of the same type. If you need something more complicated like this, you wouldn't use parsers.

luxonis_ml/data/parsers/luxonis_parser.py
```diff
@@ -111,7 +111,7 @@ def __init__(
         dataset_name = dataset_name or name.replace(" ", "_").split(".")[0]
 
         self.dataset = self.dataset_constructor(dataset_name=dataset_name, **kwargs)
-        self.parser = self.parsers[self.dataset_type](self.dataset)
+        self.parser = self.parsers[self.dataset_type](self.dataset, task_mapping or {})
```
conorsim, line 114: [Minor] Would it make sense to define the parameter as `task_mapping: Dict[LabelType, str] = {}` instead of this `or` statement?

conorsim (approved):
Left a minor comment but LGTM

luxonis_ml/data/parsers/luxonis_parser.py
kozlov721: Python can't handle well mutable default parameters (although here it probably wouldn't be an issue but having `None` there is still always the safer option).

PR #142, opened 2024-08-07
Pixel Masks Annotations

kozlov721:
- Added support for using pixel mask annotations for segmentations
  - Masks are internally transformed to RLE format

**Example:**
```python
def generator():
    mask = np.zeros((256, 256))
    yield {
        "file": file,
        "annotation": {
            "type": "mask",
            "class": "person",
            "mask": mask,
        }
    }    

```

### DIFF
--- a/luxonis_ml/data/datasets/annotation.py
+++ b/luxonis_ml/data/datasets/annotation.py
@@ -4,9 +4,10 @@ from datetime import datetime
 from typing import Any, ClassVar, Dict, List, Literal, Optional, Tuple, TypedDict, Union
 
 import numpy as np
+import numpy.typing as npt
 import pycocotools.mask as mask_util
 from PIL import Image, ImageDraw
-from pydantic import Field, model_validator
+from pydantic import ConfigDict, Field, field_validator, model_validator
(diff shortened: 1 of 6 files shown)


luxonis_ml/data/datasets/annotation.py
```diff
@@ -231,25 +254,76 @@ def get_value(self) -> Dict[str, Any]:
...
+
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _convert_rle(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        if "mask" in values:
+            return values
+
+        if "width" not in values or "height" not in values or "counts" not in values:
```
klemen1999, line 285: Why do we allow/check for "rle" type of annotations in the "mask" annotation type?

kozlov721: That's for internal conversion back from RLE. Basically, the flow is `mask (user) -> RLE (internal) -> parquet (internal) -> RLE (internal) -> mask (user)`

PR #145, opened 2024-08-14
Custom Splits

kozlov721:
- Updated `make_splits` to support custom split names instead of only `"train"`, `"val"`, and `"test"`.
  - Kept backwards compatibility with passing `ratios` as a 3-tuple for  `"train"`, `"val"`, and `"test"`
- Added tests for `make_splits`
- Fixed making new splits for a dataset that already has splits 
  - When `make_splits` is called on a dataset that already have split definitions, the old definitions stay the same and only newly added data are processed
  - Before; all the splits were generated again when `make_splits` was called

### DIFF
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -624,12 +624,13 @@ class LuxonisDataset(BaseDataset):
                     )
                 else:
                     ids = df.select("uuid").unique().get_column("uuid").to_list()
+                    old_splits = defaultdict(list)
 
             np.random.shuffle(ids)
             N = len(ids)
             lower_bound = 0
             for split, ratio in ratios.items():
-                upper_bound = lower_bound + round(N * ratio)
+                upper_bound = lower_bound + math.ceil(N * ratio)
                 new_splits[split] = ids[lower_bound:upper_bound]
                 splits_to_update.append(split)
                 lower_bound = upper_bound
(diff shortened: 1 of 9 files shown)


luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -531,36 +563,86 @@ def get_splits(self) -> Optional[Dict[str, List[str]]]:
...
+            ids = (
+                df.filter(~pl.col("uuid").is_in(defined_uuids))
+                .select("uuid")
+                .unique()
+                .get_column("uuid")
+                .to_list()
+            )
+            if not ids:
+                if not replace_old_splits:
+                    raise ValueError(
```
kozlov721, line 621: Not sure what the default behavior should be when trying to make splits with no new data. Whether to raise an error or just log a warning and do nothing.

klemen1999: I would say that this would be a "user error" and we should explicitly notify him of it as most of the time warnings aren't read.

klemen1999 (approved):
LGTM.
I would just have one more general question: Is there a point to still refer to different splits as "view" in `luxonis-train` and `luxonis-ml` `LuxonisLoader`? We refer to them as "splits" in the code (e.g. `make_splits()`,....) but when we want to access them they are called `views`. Code-wise this is the same thing so IMO it would be less confusing to also have the same name accross the codebase (also `split` is more used in other libraries/frameworks). Thoughts on unifying this? It would be a breaking change though. CC: @kozlov721 @tersekmatija @conorsim

kozlov721:
Yeah I was also thinking about this. Renaming it everywhere to `splits` would make a lot of sense. The changes required in our services wouldn't be large, but I can also keep it backwards compatible for now and still accept the `view` parameter (with a deprecation warning).

PR #147, opened 2024-08-15
Augmentation SpeedUp

sokovninn:
This PR includes:

- Augmentation and annotations loading speedup ([fix/augmentation_speedup](https://github.com/luxonis/luxonis-ml/commits/fix/augmentation_speedup) by @JSabadin )
- [opencv version 4.10.0](https://github.com/luxonis/luxonis-ml/commit/7350dd60c45c20ee722c4b0deb0a6db868390714) for faster image reading by @JSabadin 
- Reordering augmentations for faster processing

## Performance Comparison

Measured on Quadro RTX4000 (8 GB GPU memory), CUDA 11.7, 8 CPU, 32GB RAM.

Params:
dataset: COCO train split
batch_size: 16
num_workers: 4 (available with [Remove num_workers Hotfix](https://github.com/luxonis/luxonis-train/pull/58))
augs: normalization + mosaic
model: yolov6n
img_size: 416x416

| Version | Time (minutes) |
| ---------------- | -------------- |
| Unoptimized (workers=0)      | 117           |
| Optimized        | 15 :fire:            |


JSabadin (approved):

JSabadin:
I also conducted experiments to compare the performance of YOLOv6/YOLOv8 nano models with ours under the same settings and observed slightly longer timings. It took 18 minutes for our model and 21 minutes for the YOLOv6 and YOLOv8 models.

Additionally, note that previously, an epoch without the polars library, load speedup, and augmentation fixes lasted ~**5 hours** on GCP. Reducing the time to ~**15 minutes** represents a very significant speedup.

kozlov721 (approved):

luxonis_ml/data/augmentations/utils.py
```diff
@@ -173,32 +173,37 @@ def __call__(
...
+        transform_args = {
+            "image_batch": image_batch,
+            "bboxes_batch": bboxes_batch,
+            "bboxes_visibility_batch": bboxes_visibility_batch,
+            "bboxes_classes_batch": bboxes_classes_batch,
+            "keypoints_batch": keypoints_batch,
+            "keypoints_visibility_batch": keypoints_visibility_batch,
+            "keypoints_classes_batch": keypoints_classes_batch,
+        }
+        if return_mask:
```
tersekmatija, line 185: Do we see any concerns with mask_batch missing downstream? Would it be safer to keep it in dict but have it as an empty value?

JSabadin: No, the augmentations will still work if mask_batch is simply missing. I think it does not pose any threat.

 Moreover, using dummy keypoints (just one keypoint at x:0, y:0) or bounding boxes does not affect the time it takes to apply the augmentations.

sokovninn: CC: @JSabadin

PR #153, opened 2024-08-21
Layout Inference

kozlov721:
Added layout inference to NN Archive. Currently part of `modelconverter`, but I figured it makes more sense here.

### DIFF
--- a/luxonis_ml/nn_archive/config_building_blocks/base_models/output.py
+++ b/luxonis_ml/nn_archive/config_building_blocks/base_models/output.py
@@ -1,3 +1,4 @@
+from contextlib import suppress
 from typing import List, Optional
 
 from pydantic import Field, model_validator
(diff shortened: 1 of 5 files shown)


luxonis_ml/nn_archive/config_building_blocks/base_models/output.py
```diff
@@ -49,3 +51,11 @@ def validate_layout(self) -> Self:
             raise ValueError("Layout and shape must have the same length.")
 
         return self
+
+    @model_validator(mode="after")
```
klemen1999, line 55: What is the reson that for input we validate `before` and for output we validate `after`?

klemen1999 (approved):
LGTM

luxonis_ml/nn_archive/config_building_blocks/base_models/output.py
kozlov721: Output's layout is optional, so we can wait and validate after. For input it's mandatory, so it has to be validated before in order to determine whether the layout was explicitly provided (no inference in that case), or not.

PR #154, opened 2024-08-22
Augmentation SpeedUp fix

sokovninn:
This commit includes:

- `Resize()` after all other augmentations to ensure the correct input shape
- Optimizing amount of `Resize()` calls

Some benchmarks on the COCO train dataset (1 epoch, 4 workers, bs 16):

- Mosaic + set of other augmentations (do change shape):
  - resize before augs + resize if needed after augs: 22:26
  - resize after augs: 20:41
  - resize before normalization only + resize if needed after augs: 20:57 - chosen solution

- Only mosaic augmentation (doesn't change shape):
  - resize before augs + resize if needed after augs: 14:53
  - resize after augs: 15:42
  - resize before normalization only: 15:24
  - resize before normalization only + resize if needed after augs: 14:49 - chosen solution


luxonis_ml/data/augmentations/utils.py
```diff
@@ -114,7 +116,19 @@ def _parse_cfg(
...
+                format="coco", label_fields=["bboxes_classes", "bboxes_visibility"]
+            ),
+            keypoint_params=A.KeypointParams(
+                format="xy",
+                label_fields=["keypoints_visibility", "keypoints_classes"],
+                remove_invisible=False,
+            ),
+        )
+
+        return batch_transform, spatial_transform, resize_transform
```
kozlov721, line 131: The annotation of `_pase_cfg` is outdated

kozlov721 (approved):
LGTM

luxonis_ml/data/augmentations/utils.py
```diff
@@ -114,7 +118,60 @@ def _parse_cfg(
...
+                format="xy",
+                label_fields=["keypoints_visibility", "keypoints_classes"],
+                remove_invisible=False,
+            ),
+        )
+
+        return batch_transform, spatial_transform, pixel_transform, resize_transform
+
+    def _apply_transform(
+        self, transformed, arg_names, transform_func, return_mask=False, arg_suffix=""
```
klemen1999, line 148: Let's add type annotations here

luxonis_ml/data/augmentations/utils.py
```diff
@@ -299,21 +386,30 @@ def post_transform_process(
...
-                out_mask[int(key) - 1, ...] = transformed_mask == key
-        out_mask[out_mask > 0] = 1
+        out_image = out_image.astype(np.float32)
+
+        out_mask = None
+        if return_mask:
+            transformed_mask = transformed_data.get("mask")
+            out_mask = (
+                np.zeros((ns, *transformed_mask.shape))
+                if transformed_mask is not None
```
klemen1999, line 404: If `return_mask=True` in which case would we get that `transformed_mask=None` ?

luxonis_ml/data/augmentations/utils.py
```diff
@@ -287,7 +373,8 @@ def post_transform_process(
         ns: int,
         nk: int,
         filter_kpts_by_bbox: bool,
-    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        return_mask: bool = True,
+    ) -> Tuple[np.ndarray, Optional[np.ndarray], np.ndarray, np.ndarray]:
```
klemen1999, line 377: From type complience perspective should we make default `out_mask=np.zeros(...)` instead of `None` since `__call__` should return `Dict[LabelType, np.ndarray]`? cc: @kozlov721

klemen1999 (commented):
Left some comments otherwise LGTM

luxonis_ml/data/augmentations/utils.py
sokovninn: @JSabadin could you answer this?

luxonis_ml/data/augmentations/utils.py
sokovninn: We don't return mask if it's `None`.

klemen1999: Where don't we return it? Because we return it from `post_transform_process` as `None` and then this is part of out_annotations in `__call__`'s return right?

sokovninn: If mask is in `present_annotations` we return it. Otherwise we return None from `post_transform_process` and nothing from `__call__`. CC: @JSabadin to confirm

klemen1999 (approved):

PR #168, opened 2024-09-11
Duplicate Data Warning

JSabadin:
This PR introduces the `warn_on_duplicates `method, which checks for and logs:

1. Files with the same UUID but different file names.
2. Files with duplicate annotations.

It efficiently detects these cases and raises warnings to help identify issues in the dataset.

### DIFF
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -154,15 +154,20 @@ class LuxonisDataset(BaseDataset):
                 self.bucket_storage, self.bucket, self.team_id, self.dataset_name
             )
 
-    def _load_df_offline(self) -> Optional[pl.DataFrame]:
+    def _load_df_offline(
+        self, lazy: bool = False
+    ) -> Optional[pl.LazyFrame | pl.DataFrame]:
         path = get_dir(self.fs, "annotations", self.local_path)
 
         if path is None or not path.exists():
             return None
 
-        dfs = [pl.read_parquet(file) for file in path.glob("*.parquet")]
-
-        return pl.concat(dfs) if dfs else None
+        if lazy:
+            dfs = [pl.scan_parquet(file) for file in path.glob("*.parquet")]
+            return pl.concat(dfs) if dfs else None
+        else:
+            dfs = [pl.read_parquet(file) for file in path.glob("*.parquet")]
+            return pl.concat(dfs) if dfs else None
 
     def _get_file_index(self) -> Optional[pl.DataFrame]:
         path = get_file(self.fs, "metadata/file_index.parquet", self.media_path)
@@ -503,12 +508,11 @@ class LuxonisDataset(BaseDataset):
 
         self.fs.put_file(tmp_file.name, "metadata/file_index.parquet")
         self._write_metadata()
-        self.warn_on_duplicates()
+        self._warn_on_duplicates()
         return self
 
-    def warn_on_duplicates(self) -> None:
-        df = self._load_df_offline()
-
+    def _warn_on_duplicates(self) -> None:
+        df = self._load_df_offline(lazy=True)
         # Warn on duplicate UUIDs
         duplicates_paired = (
             df.group_by("uuid")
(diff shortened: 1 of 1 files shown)


luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -503,9 +503,38 @@ def add(self, generator: DatasetIterator, batch_size: int = 1_000_000) -> Self:
 
         self.fs.put_file(tmp_file.name, "metadata/file_index.parquet")
         self._write_metadata()
-
+        self.warn_on_duplicates()
         return self
 
+    def warn_on_duplicates(self) -> None:
```
kozlov721, line 509: I think this should be a private method

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -503,9 +503,38 @@ def add(self, generator: DatasetIterator, batch_size: int = 1_000_000) -> Self:
...
+        # Warn on duplicate UUIDs
+        duplicates_paired = (
+            df.group_by("uuid")
+            .agg(pl.col("file").n_unique().alias("file_count"))
+            .filter(pl.col("file_count") > 1)
+            .join(df, on="uuid")
+            .select(["uuid", "file"])
+            .unique()
+            .group_by("uuid")
+            .agg([pl.col("file").alias("files")])
```
kozlov721, line 521: Could this be further extended with `pl.len` and a filter for len > 1 so we don't have to do it in `iterrows`? It should make it faster. 
Not super important since this shouldn't be a bottleneck, but I'd assume issues like this are more likely the larger the dataset is, so it would be good to have this optimized.

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -503,9 +503,38 @@ def add(self, generator: DatasetIterator, batch_size: int = 1_000_000) -> Self:
 
         self.fs.put_file(tmp_file.name, "metadata/file_index.parquet")
         self._write_metadata()
-
+        self.warn_on_duplicates()
         return self
 
+    def warn_on_duplicates(self) -> None:
+        df = self._load_df_offline()
```
kozlov721, line 510: Here we could make use of a lazy data frame for a better performance. Could you modify the `_load_df_offline` so it takes `lazy: bool = False` and returns a lazy frame if set to `True`? Check [`pl.scan_parquet`](https://docs.pola.rs/api/python/stable/reference/api/polars.scan_parquet.html).

kozlov721 (commented):
LGTM, just left a few performance-oriented comments. Not super important, if we need this PR merged quickly, we can add the optimizations in a different PR.

luxonis_ml/data/datasets/luxonis_dataset.py
JSabadin: Made changes in the new commit.

luxonis_ml/data/datasets/luxonis_dataset.py
JSabadin: Made changes in the new commit

luxonis_ml/data/datasets/luxonis_dataset.py
JSabadin: Made changes in the new commit

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -154,15 +154,20 @@ def _init_paths(self) -> None:
                 self.bucket_storage, self.bucket, self.team_id, self.dataset_name
             )
 
-    def _load_df_offline(self) -> Optional[pl.DataFrame]:
+    def _load_df_offline(
+        self, lazy: bool = False
+    ) -> Optional[pl.LazyFrame | pl.DataFrame]:
```
klemen1999, line 159: Should use `Union` for python3.8 tests to pass: `Optional[Union[pl.LazyFrame,pl.DataFrame]]`

klemen1999 (approved):
LGTM

luxonis_ml/data/datasets/luxonis_dataset.py
JSabadin: Made changes in the '_Improved type annotations_' commit.

kozlov721 (approved):
LGTM

PR #176, opened 2024-10-01
[Feat] Added name attribute to HeadType

klemen1999:
Added optional `name` parameter to Head in NNArchive definition. This can be used for easier postprocessing in DAI/depthai-nodes

### DIFF
--- a/luxonis_ml/nn_archive/config.py
+++ b/luxonis_ml/nn_archive/config.py
@@ -6,7 +6,7 @@ from luxonis_ml.utils import BaseModelExtraForbid
 
 from .model import Model
 
-CONFIG_VERSION = Literal["1.1"]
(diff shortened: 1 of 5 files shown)


kozlov721 (approved):
LGTM

luxonis_ml/nn_archive/config.py
```diff
@@ -6,7 +6,7 @@
 
 from .model import Model
 
-CONFIG_VERSION = Literal["1.0"]
+CONFIG_VERSION = Literal["1.1"]
```
kozlov721, line 9: This should be changed it to `Literal["1.0", "1.1"]`. Otherwise, we won't be able to load old NN Archive configs after this is merged.

PR #177, opened 2024-10-02
Added automatic clipping for normalized annotations

klemen1999:
Now values that should be normalized in [0,1] range for annotations are automatically clipped to this if not already. 
Reason behind it is that even official COCO dataset includes values that are outside of [0,1] range when automatically normalized and thus the automatic parsing fails. If automatic clipping is added then we save user the effort to manually go through all annotations, find specific annotation where this happens and fix it.
Obvious downside is that now the validation would pass even if user doesn't actually pass normalized values because they would be clipped automatically to correct range - let me know if you have a better solution for this.

But IMO the upside is greater because automatic parsing failing just because one value is 1.001 instead of 1 is quite annoying.

### DIFF
--- a/luxonis_ml/data/datasets/annotation.py
+++ b/luxonis_ml/data/datasets/annotation.py
@@ -145,12 +145,17 @@ class BBoxAnnotation(Annotation):
     def validate_values(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         warn = False
         for key in ["x", "y", "w", "h"]:
+            if values[key] < -2 or values[key] > 2:
+                raise ValueError(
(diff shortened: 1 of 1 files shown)


klemen1999:
Note: Tests are failing because of onnx version (fixed in #176 ) because yesterday onnx==1.17.0 came out and it looks like there are some breaking changes.

tersekmatija:
> Obvious downside is that now the validation would pass even if user doesn't actually pass normalized values because they would be clipped automatically to correct range

Could we raise a warning or fail if value would be `> 2.0`? This would mean that annotations are unlikely normalized, while small "overflows" like `1.002` would be still correctly clipped.

kozlov721 (approved):
LGTM

PR #183, opened 2024-10-03
Speedup in picking batched aug indices

klemen1999:
Previous method for picking random indices for batched augs included creating a list with all possible indices. This can get slow when datasets are large since it happens on every getitem() call.
If the dataset size is smaller than size of batches we can still do it this way since it means that the dataset is of max size 4.

But if this is not the case then it's better to just pick random values in the range until we have enought of them. The speedup is seen on the image bellow comparing times for picking 4 random indices from datasets of different sizes (times averaged across  10 runs).
comparison


kozlov721 (approved):
LGTM

PR #184, opened 2024-10-04
Improved Duplicate Warnings

kozlov721:
- Improved duplicate warnings
  - The warnings are not raised on empty annotations, annotations of different tasks and of different instance ids
  - The warning message contain the absolute file path instead of just the file name
  - Added the task and annotation type to the warning message
  - The content of RLE and mask segmentation annotations is not logged 
- Added option to lazy load file index
- Added path resolving (removing `.` and `..` directories from path)


klemen1999 (approved):
LGTM thanks, previous warnings were a bit excessive

PR #190, opened 2024-10-16
[Hotfix] Config version set to 1.0

klemen1999:
Changed config version from `1.1` back to `1.0`. 
Version was previously bumped to 1.1 because we added optional `name` parameter to the `head`. But since this parameter is optional this is not a breaking change and we should leave the version of the config at 1.0 so it doesn't interfere with the schema validation done on HubAI or by DepthAI.

I also added a test which checks if the validation goes through if `name` field is specified and if it isn't. The E2E workflow (luxonis-train -> archive -> upload to HubAI -> convert -> run on device) was verified and works as expected.


tersekmatija (approved):

kozlov721 (approved):

moratom (approved):
LGTM

conorsim (approved):

PR #193, opened 2024-10-30
Fix: Prevent training from crashing when unable to connect to MLflow.

JSabadin:
# Robust MLflow Connection Handling

## Summary
This update resolves an issue where training would crash if a connection to MLflow could not be established. With this fix, training continues even if MLflow connectivity is intermittent or unavailable. Data is temporarily stored locally in RAM when a connection is not present and is uploaded once connectivity is restored.

## Tested Scenarios
This fix was validated under the following conditions:
1. Training begins without MLflow connectivity, which is then established mid-training.
2. MLflow connection is lost mid-training but reestablished later.
3. MLflow connection is lost mid-training and does not return.
4. Training runs without MLflow connectivity throughout.

## Implementation Details
- **Local Storage**: When a connection to MLflow is unavailable, log data is temporarily stored in RAM.
- **Automatic Upload**: Once MLflow connectivity is restored, all stored logs are uploaded.
- **End-of-Training Behavior**: If training ends while disconnected from MLflow, any unsent logs are saved locally to ensure data is not lost.

---

### DIFF
--- a/luxonis_ml/tracker/tracker.py
+++ b/luxonis_ml/tracker/tracker.py
@@ -1,4 +1,5 @@
 import glob
+import json
 import logging
 import os
 from functools import wraps
@@ -6,6 +7,7 @@ from importlib.util import find_spec
 from pathlib import Path
 from typing import Any, Callable, Dict, Literal, Optional, Union
 
+import cv2
 import numpy as np
 from unique_names_generator import get_random_name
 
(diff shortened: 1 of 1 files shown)


kozlov721 (approved):
LGTM

klemen1999 (approved):

PR #194, opened 2024-10-30
fix: sort classes alphabetically

JSabadin:
## Description

This PR sorts the classes alphabetically, ensuring that dataset creation is independent of the yielding order. Classes are now ordered alphabetically, providing consistency across different runs.


luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -310,6 +310,8 @@ def get_classes(self) -> Tuple[List[str], Dict[str, List[str]]]:
                 for c in classes
             }
         )
+        # for task in self.metadata["classes"]:
```
klemen1999, line 313: I believe this shouldn't be commented out right?

klemen1999 (approved):

PR #199, opened 2024-11-04
hotifx: adding file names to be ignored by the COCO parser

JSabadin:
Hotfix to ignore files that are in the training split and have exact copies with different names in the validation split

### DIFF
--- a/luxonis_ml/data/parsers/coco_parser.py
+++ b/luxonis_ml/data/parsers/coco_parser.py
@@ -136,9 +136,10 @@ class COCOParser(BaseParser):
             and dir_format == Format.FIFTYONE
             else train_paths["annotation_path"]
         )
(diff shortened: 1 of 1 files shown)


luxonis_ml/data/parsers/coco_parser.py
```diff
@@ -316,3 +320,45 @@ def generator() -> DatasetIterator:
...
+        "000000373199.jpg",
+        "000000410810.jpg",
+        "000000397819.jpg",
+        "000000578492.jpg",
+        "000000531721.jpg",
+    ]
+    with open(annotation_path, "r") as f:
+        annotation_data = json.load(f)
+
+    filtered_images = [
```
klemen1999, line 345: Can we make it so that if none of these images are preset then we don't create `labels_fixed.json` and just return the original  `annotation_path`

klemen1999 (approved):

PR #210, opened 2024-11-27
loading with empty annotations

JSabadin:
## Feature: Support for Images Without Annotations

This update adds support for loading images without annotations. When no annotations are present, empty arrays are initialized for each task (e.g., segmentation, bounding boxes, keypoints, and classification). 

>Please note that loading non-annotated images is only supported for basic tasks such as `segmentation`, `classification`, `boundingbox`, and `keypoints`. It is not supported for custom tasks.

### DIFF
--- a/luxonis_ml/data/loaders/luxonis_loader.py
+++ b/luxonis_ml/data/loaders/luxonis_loader.py
@@ -323,4 +323,21 @@ class LuxonisLoader(BaseLoader):
 
             labels[task] = (array, anns[0]._label_type)
 
+        if not labels:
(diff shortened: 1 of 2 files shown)


kozlov721 (approved):
LGTM, but please add tests for a dataset without labels.

JSabadin:
@klemen1999 @kozlov721 Please note that loading non-annotated images is only supported for basic tasks such as `segmentation`, `classification`, `boundingbox`,  and `keypoints`. It is not supported for custom tasks.

PR #211, opened 2024-12-03
Nested Labels and Augmentation

kozlov721:
# Labels and Augmentations refactor

- [Nested Labels](#nested-labels)
- [Augmentations](#augmentations)
- [Miscellaneous](#miscellaneous)
- [Breaking Changes](#breaking-changes)
   * [Migration](#migration)

## Nested Labels

- Added native support for arbitrarily nested labels 
- Added new `"metadata"` field that can hold additional information about the label
  - Can be `int`, `float`, or `string`
- `"task"` moved up from the `"annotation"` section
  - An image without labels still belongs to some task 
- Groundwork for future support of multiple images per annotation 
- Added support for instance segmentation tasks
  - New `"instance_segmentation"` task type 
- Mostly backwards compatible with LDF `v1.0.0`, no need for manual migration 
  - However, for full compatibility the dataset has to be re-created using the latest `luxonis-ml`  

**Example**
```python
yield {
    "file": "img.png",
    "task": "vehicles",
    "annotation": {
        "class": "car",
        "boundingbox": {
            "x": x, "y": y, "w": w, "h": h
        },
        "keypoints": {
            "keypoints": [...]
        },
        "instance_segmentation": {
            "mask": ...
        },
        "array": {"path": ...},
        "metadata": {
            "id": 12,
        },
        "sub_detections": {
            "license_plate": {
                "boundingbox": {...},
                "metadata": {
                    "text": "123ABC"
                },
            }
        }
    }
}
```

The output of the loader for the above annotations would be:

```python
{
    "vehicles/boundingbox": ...,
    "vehicles/keypoints": ...,
    "vehicles/instance_segmentation": ...,
    "vehicles/classification": ...,
    "vehicles/array": ...,
    "vehicles/metadata/id": ...,
    "vehicles/license_plate/boundingbox": ...,
    "vehicles/license_plate/metadata/text": ...,
}
```

## Augmentations

- Upgraded `albumentations` from `1.3.x` to `1.4.x`
- Added support for multiple batched augmentations 
  - e.g. `MixUp` together with `Mosaic4`
- Improved `MixUp` augmentation
  - Uses letterbox resizing by default
    - Can be configured by `keep_aspect_ration` parameter 
  - Improved augmenting semantic segmentations
    - Masks are blended together, in case of a conflict, the mask associated with higher $\alpha$ takes precedence
- Native support for labels with multiple annotations of the same type
  - Instead of it being a hack on the `LuxonisLoader` part   
- Cleaned up and removed unused code
- Overall performance increase for all types of augmentations (see the table below)
- Augmentations are now initialized by `BaseLoader` instead of by the user
  - This allows for augmenting all annotations in one go using albumentation's `additional_targets` logic
  - Simplified usage for users
  -`Augmentations` renamed to `AlbumentationsEngine` and inherit from `AugmentationEngine`
    - Groundwork for future support for more augmentation engines
    - Users can implement their own engine which registers into `AUGMENTATION_ENGINES` registry

**Example**
**`old.py`**
```python
from luxonis_ml.data import LuxonisDataset, LuxonisLoader, Augmentations


dataset = LuxonisDataset(...)
augmentations = Augmentations((height, width), config, ...)
loader = LuxonisLoader(dataset, augmentations, ...)
```

**`new.py`**
```python
from luxonis_ml.data import LuxonisDataset, LuxonisLoader

dataset = LuxonisDataset(...)
loader = LuxonisLoader(
    dataset, 
    height=...,
    width=...,
    # possible future support for other engines, e.g. Nvidia's DALI
    augmentation_engine="albumentations",
    augmentation_config=...,
    ...,
)
```

**Benchmark**

Benchmarking was done on [Microsoft COCO 2017 Dataset](https://public.roboflow.com/object-detection/microsoft-coco-subset/2). The size of the dataset is 92,325 images.
The new augmentations are faster in all categories.

| Action | Old | New |
|--------|--------|--------|
| Dataset Creation | 118s | 112s |
| Loader Init | 184s | 138s |
| Iterating No Augmentation | 91s | 89s |
| Iterating Standard Augmentations | 620s | 582s |
| Iterating MixUp Augmentation | 401s | 307s |
| Iterating Mosaic4 Augmentation | 590s | 505s | 

## Miscellaneous

- Improved tests
  - 96 % coverage 
- Added new directory `tools` to the repository that will contain various scripts for easier development
  - Currently contains `benchmark.py` 
- Added new abstract property `version` to `BaseDataset`
- Added new `luxonis_ml checkhealth` command that prints a table with the state of individual sub-packages including possible import errors
- Custom types moved to `luxonis_ml.typing`
- Improved visualization functions for `luxonis_ml data inspect`
  - Nested labels rendered on top of each other
  - Occluded keypoints marked with a cross
  - Nested bounding boxes rendered using dashed lines 
  - Added `--keep_aspect_ratio` (or `-k`) flag to use letterbox resizing with augmentations
  - Added `--blend-all` (or `-b`) flag to draw all labels on top of the same image (disregarding different tasks)
image


luxonis_ml/data/datasets/annotation.py
```diff
@@ -505,114 +442,83 @@ class ArrayAnnotation(Annotation):
...
-    """
+    @model_validator(mode="before")
+    @classmethod
+    def validate_task_name(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        if "task_name" not in values and "task" not in values:
+            if "segmentation" in values.get("annotation", {}):
+                values["task_name"] = "segmentation"
+            else:
+                values["task_name"] = "detection"
+        return values
```
klemen1999, line 494: Is this us trying to automaticaly set task_name if not present based on "guessing" depending on what other fields are present? Should we rather make this field required?
Also would now be a good time to pick either only "task" or "task_name" and not support both going forward?

luxonis_ml/data/datasets/annotation.py
```diff
@@ -2,142 +2,151 @@
...
+                "class_name": self.class_name,
+                "instance_id": self.instance_id,
+                "label_type": f"{prefix}classification",
+                "annotation": "{}",
+            }
+        for name, detection in self.sub_detections.items():
+            yield from detection.to_parquet_rows(f"{prefix}{name}/")
+
+    @model_validator(mode="after")
+    def rescale_values(self) -> Self:
```
klemen1999, line 102: When is this used?

luxonis_ml/data/parsers/coco_parser.py
```diff
@@ -241,17 +241,10 @@ def generator() -> DatasetIterator:
...
-                    yield {
-                        "file": path,
-                        "annotation": {
-                            "type": "classification",
-                            "class": class_name,
-                        },
-                    }
 
                     seg = ann["segmentation"]
                     if isinstance(seg, list):
```
klemen1999, line 246: We should also merge segmentation labels into unified "record" dict and yield together with other labels (class, bbox and kpts) otherwise now we can't parse COCO to include all 4 type of labels

luxonis_ml/data/parsers/solo_parser.py
```diff
@@ -158,136 +158,109 @@ def from_split(
...
+
+                            for instance in sseg_annotations["instances"]:
+                                class_name = instance["labelName"]
+                                r, g, b, _ = instance["pixelValue"]
+                                curr_mask = np.zeros_like(mask)
+                                curr_mask[
+                                    np.all(mask == [b, g, r], axis=2)
+                                ] = 1
+                                curr_mask = np.max(curr_mask, axis=2)  # 3D->2D
+                                yield {
```
klemen1999, line 198: Should include in the yield together with keypoints below?

klemen1999 (commented):
Generally LGTM, left some comments/questions.
Also before merge we shuold update the  README in "data" dir and add an automatic conversion function for older datasets to port them to new one.

luxonis_ml/data/augmentations/custom/letterbox_resize.py
```diff
@@ -1,23 +1,20 @@
...
-class LetterboxResize(DualTransform):
+class LetterboxResize(A.DualTransform):
     def __init__(
         self,
         height: int,
         width: int,
         interpolation: int = cv2.INTER_LINEAR,
         border_value: int = 0,
         mask_value: int = 0,
-        always_apply: bool = False,
```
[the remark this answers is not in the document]
kozlov721: Yeah

luxonis_ml/data/augmentations/custom/mixup.py
```diff
@@ -118,61 +95,51 @@ def apply_to_mask_batch(
...
-        @param kwargs: Additional parameters for the transformation.
-        @rtype: List[BoxType]
-        @return: List of transformed bboxes.
+        @rtype: np.ndarray
+        @return: Transformed bboxes.
         """
-        return [bboxes_batch[0] + bboxes_batch[1]]
+        for i in range(len(bboxes_batch)):
+            bbox = bboxes_batch[i]
+            if bbox.shape[1] == 4:
```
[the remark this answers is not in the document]
kozlov721: When all boxes are filtered out. The empty array that is returned if no boxes are present is (0,4), but the existing boxes are of shape (N,6).

luxonis_ml/data/datasets/annotation.py
kozlov721: It's used if you are using nested labels and want to pass for example the keypoints using coordinates relative to the bbox and not the entire image. I'm not sure though if there's ever going to be a use case for this, so maybe we can just remove this option.

luxonis_ml/data/parsers/coco_parser.py
kozlov721: The COCO is still parseable. The reason it's not in the same record is because the segmentation masks are not instance segmentation, but rather one mask for all the people. So it's assigned its own task instead of it being added to `detection` task.

luxonis_ml/data/parsers/solo_parser.py
kozlov721: The same situation like COCO above.

luxonis_ml/data/datasets/annotation.py
kozlov721: Yeah. We need to somehow keep in mind that segmentation is usually not instance based so it has to be its own task. 
We can make the task required. It just wasn't before so I didn't want to change it. Also usually you're only using datasets with a single task so it's nice that you don't have to think of tasks in that case.

luxonis_ml/data/augmentations/augmentations.py
```diff
@@ -0,0 +1,377 @@
...
+            data["keypoints_visibility"],
+            data["keypoints_classes"],
+            n_keypoints,
+            image_height,
+            image_width,
+        )
+
+        visible_bboxes = [int(v) for v in data["bboxes_visibility"]]
+
+        if {"boundingbox", "keypoints"} < present_labels:
```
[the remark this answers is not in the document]
klemen1999: But if `present_labels = {"boundingbox", "keypoints"}` this now resolves to False. Is this wanted?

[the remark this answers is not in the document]
kozlov721: Oh I see, you're right. That's not wanted

luxonis_ml/data/datasets/annotation.py
klemen1999: Hm yeah agree that it is nice when it autopopulates since most of the time there is only one task. I fine with it keeping it not required if we are able to infer this value reliably. Or if it is ambiguous then warn/error out (probably want to error in this case)

luxonis_ml/data/datasets/annotation.py
klemen1999: I see. I guess we can keep then for this option to be available

luxonis_ml/data/parsers/coco_parser.py
kozlov721: The label keys would be 
```python
[
    'segmentation/segmentation',
    'segmentation/classification',
    'detection/boundingbox',
    'detection/keypoints',
    'detection/classification'
]
```

Slight annoyance is the duplicate `classification` labels for both tasks, but I'm not sure if we can easily avoid it as the two tasks can indeed have completely different classes. We just know they don't in this case.

kozlov721: Actually I just figured out how we can easily fix this. We just need to explicitly set the `"task"` to the same value for both the segmentation and detection records.

kozlov721:
**Augmentation Update**

Moved the initialization of augmentations from user side to the loader. 

**Reasoning**

This change will make it possible to augment all the annotations in one go by passing the extra ones as `additional_targets`. The additional targets must be defined at `A.Compose`'s initialization time so it needs to be done by the loader. 

As a bonus it also simplifies usage for the user.

**Example**
**`old.py`**
```python
from luxonis_ml.data import LuxonisDataset, LuxonisLoader, Augmentations


dataset = LuxonisDataset(...)
augmentations = Augnemtations.from_config(height, width, config, ...)
loader = LuxonisLoader(dataset, augmentations, ...)
```

**`new.py`**
```python
from luxonis_ml.data import LuxonisDataset, LuxonisLoader

dataset = LuxonisDataset(...)
loader = LuxonisLoader(
    dataset, 
    height=...,
    width=...,
    # possible future support for other engines, e.g. Nvidia's DALI
    augmentation_engine="albumentations",
    augmentation_config=...,
    ...,
)
```

luxonis_ml/data/__main__.py
```diff
@@ -194,30 +187,25 @@ def inspect(
...
-        )
 
     if len(dataset) == 0:
         raise ValueError(f"Dataset '{name}' is empty.")
 
-    loader = LuxonisLoader(dataset, view=view, augmentations=augmentations)
+    loader = LuxonisLoader(
+        dataset,
+        view=view,
+        augmentation_config=aug_config,
```
klemen1999, line 204: We want to document well how should this `augmentation_config` look like

luxonis_ml/data/augmentations/custom/mixup.py
```diff
@@ -101,7 +101,7 @@ def apply_to_mask_batch(
         return out_mask
 
     @override
-    def apply_to_bboxes_batch(
+    def apply_to_bboxes(
```
klemen1999, line 104: Is now batched augmentation transformed to the point that it can call the same API as the normal (1 image) augmentations? As I see most of the batch reated code got removed and also this now overrides the methods of single image augmentations.

luxonis_ml/data/augmentations/augmentations.py
```diff
@@ -85,41 +77,34 @@ def from_config(
...
-            suffix = "_batch" if batch else ""
+        def _get_params():
             return {
                 "bbox_params": A.BboxParams(
-                    format="coco",
-                    label_fields=[
-                        f"bboxes_classes{suffix}",
-                        f"bboxes_visibility{suffix}",
-                    ],
+                    format="albumentations",
```
klemen1999, line 89: Any specific reason for switching to albumentations format from coco?

luxonis_ml/data/augmentations/augmentations.py
```diff
@@ -85,41 +77,34 @@ def from_config(
...
+                    label_fields=["bboxes_classes", "bboxes_visibility"],
+                    # Bug in albumentations v1.4.18 (the latest installable
+                    # on python 3.8) causes the pipeline to eventually
+                    # crash when set to 0.
```
klemen1999, line 93: When what is set to 0? Do we want to exclude this version from requrements if we know it causes problems?

luxonis_ml/data/datasets/annotation.py
```diff
@@ -468,7 +461,7 @@ def serialize_path(self, value: FilePath) -> str:
 class DatasetRecord(BaseModelExtraForbid):
     files: Dict[str, FilePath]
     annotation: Optional[Detection] = None
-    task_name: str = Field(validation_alias=AliasChoices("task", "task_name"))
+    task: str
```
klemen1999, line 464: If I understand correctly now the key is "task" right? If possible I would then internally also stick to this name for the variable names. It gets a bit confusing when sometimes we call it "task" and sometimes "task_name"  (e.g. in `def to parquet_rows()` and also other places)

luxonis_ml/data/augmentations/augmentations.py
kozlov721: It's 0 by default (bbox must be completely outside of the image to be excluded), but for some reason this is buggy up until and including `1.4.18`. Unfortunately, this is the latest version available for Python 3.8.

luxonis_ml/data/augmentations/augmentations.py
kozlov721: Less pre- and post-processing

luxonis_ml/data/augmentations/custom/mixup.py
kozlov721: Yeah, it uses the same API now. Most of the batch related code wasn't necessary for our usage.

luxonis_ml/data/datasets/annotation.py
kozlov721: I'm trying to keep this convention:
- `task` (with the single exception in the record for better backwards compatibility) means the full task string: `"detection/segmentation"`
- `task_name` refers only to the first part: `"detection"`

luxonis_ml/data/augmentations/augmentations.py
kozlov721: I'd argue this is virtually correct. But we can make the number even smaller.

klemen1999: Yeah if the issue is only the exact 0 and some float close to 0 doesn't cause it then I think it's fine to set it like that. And make a comment why this is like that so we remember not to change it

luxonis_ml/data/augmentations/augmentations.py
```diff
@@ -0,0 +1,359 @@
...
+            batch_data: List[LoaderOutput] = []
+            nk = 0
+            ns = 0
+            for img, labels in data:
+                label_subset = {}
+                for task, label in labels.items():
+                    if get_qualified_task_name(task) == task_name:
+                        task_type = get_task_type(task)
+                        if task_type == "instance_segmentation":
+                            task_type = "segmentation"
```
sokovninn, line 144: Why the task_type is changed here?

luxonis_ml/data/augmentations/augmentations.py
```diff
@@ -0,0 +1,359 @@
...
+            data["bboxes"], np.expand_dims(data["bboxes_classes"], axis=-1)
+        )
+        out_keypoints = post_process_keypoints(
+            data["keypoints"],
+            data["keypoints_visibility"],
+            data["keypoints_classes"],
+            n_keypoints,
+            image_height,
+            image_width,
+        )
```
sokovninn, line 327: Are invisible keypoints removed, or are their values just set to 0?

sokovninn (approved):
LGTM, great job! Left a few minor comments.

luxonis_ml/data/augmentations/augmentations.py
kozlov721: From albumentation's perspective instance segmentation is treated the same as normal segmentation.

luxonis_ml/data/augmentations/augmentations.py
kozlov721: Values are set to 0 as per coco convention.

sokovninn: So zeros for invisible keypoints present in the targets during training?

kozlov721: Yeah. The model predicts fixed number of keypoints so we can't remove the invisible ones. Instead, they are set to 0.

sokovninn: I see, thanks!

JSabadin (approved):
LGTM! An additional table comparing image loading with augmentations before and after the update would be great.

kozlov721:
**Update with benchmarks**

Benchmarking was done on [Microsoft COCO 2017 Dataset](https://public.roboflow.com/object-detection/microsoft-coco-subset/2). The size of the dataset is 92,325 images.
The new augmentations are faster in all categories.

| Action | Old | New |
|--------|--------|--------|
| Dataset Creation | 118s | 112s |
| Loader Init | 192s | 138s |
| Iterating No Augmentation | 91s | 89s |
| Iterating Standard Augmentations | 620s | 582s |
| Iterating MixUp Augmentation | 401s | 307s |
| Iterating Mosaic4 Augmentation | 590s | 505s |

tools/migrate.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for dataset_name in datasets:
+        base_path = (
+            environ.LUXONISML_BASE_PATH
+            / "data"
+            / team_id
+            / "datasets"
+            / dataset_name
+        )
+        for parquet_file in (base_path / "annotations").glob("*.parquet"):
```
klemen1999, line 61: Can we check LDF_VERSION of the dataset and migrate with this migrate function only if previous one is e.g. 1.0? Just so we "future proof" this tool so when e.g. LDF 3.0 goes out we'll probably have different migrate function from 1.0->3.0 and different from 1.0->2.0 or 2.0->3.0

klemen1999 (approved):
LGTM, thanks!
Let's fix the tests and then we can merge IMO

tools/migrate.py
klemen1999: Yea this is more a suggestion, not needed at this moment.

PR #215, opened 2024-12-04
Fix overlapping masks in semantic segmentation targets

sokovninn:
This PR resolves the issue of overlapping masks in semantic segmentation targets.

### Benchmark Results (1 epoch COCO, 5000 images, 8 workers):

- **Before**: 4:00 minutes
- **After**: 4:13 minutes (approximately 6% increase in processing time)

### Notes:
The added overhead is due to the additional processing required to remove overlaps in the segmentation masks. Further optimizations may be explored if necessary.

### Update:
After further optimizations there is no overhead anymore.


kozlov721:
> (approximately 6% increase in processing time)
IMHO the overhead is too big. We should check and correct this on the side of `LuxonisDataset` when the dataset is being created.

kozlov721 (approved):
Thanks, LGTM! Could you just add a small test for this?

PR #220, opened 2025-01-06
feat: skip dataset re-download and ensure safe dataset syncing

JSabadin:
### Skip Redownload Dataset and Fix Potential Race Conditions

A new parameter, `update_mode`, has been introduced in the `LuxonisLoader`. This parameter allows more control over when the dataset from the cloud is already downloaded locally.

- If set to `UpdateMode.ALWAYS`, the loader will always re-download the dataset locally.
- If set to `UpdateMode.IF_EMPTY`, it will only download the dataset if it does not already exist locally.

**Example Usage:**

```python
from luxonis_loader import LuxonisLoader, UpdateMode

loader = LuxonisLoader(
    dataset,
    update_mode=UpdateMode.IF_EMPTY
)
```

This ensures that the loader uses the local dataset if one is already present, helping to avoid unnecessary downloads.

---

### FileLock Added for Safe Syncing

To address potential race conditions, a `FileLock` mechanism has been implemented in the following areas:

1. **`sync_from_cloud` Method:** This prevents multiple processes from attempting to sync the dataset simultaneously in a distributed environment (e.g., DDP on GCP).
2. **`_get_metadata` in Dataset Initialization:** Ensures safe concurrent access when multiple processes initialize the dataset and loader before setting up the DDP environment.

Additionally, the `_load_df_offline` method has been updated to:

- Only access local storage.
- Avoid attempting to download dataset annotations from the cloud, as `sync_from_cloud` already handles this.

These changes help to ensure safe and efficient operation in distributed environments by preventing redundant downloads and race conditions.

### DIFF
--- a/luxonis_ml/data/loaders/luxonis_loader.py
+++ b/luxonis_ml/data/loaders/luxonis_loader.py
@@ -18,6 +18,7 @@ from luxonis_ml.data.augmentations import (
 from luxonis_ml.data.datasets import (
     Annotation,
     LuxonisDataset,
+    UpdateMode,
     load_annotation,
 )
 from luxonis_ml.data.loaders.base_loader import BaseLoader
@@ -48,8 +49,7 @@ class LuxonisLoader(BaseLoader):
         keep_aspect_ratio: bool = True,
         out_image_format: Literal["RGB", "BGR"] = "RGB",
         *,
-        force_resync: bool = False,
-        skip_redownload_dataset: bool = False,
+        update_mode: UpdateMode = UpdateMode.ALWAYS,
     ) -> None:
         """A loader class used for loading data from L{LuxonisDataset}.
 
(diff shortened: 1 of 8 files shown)


kozlov721:
Doesn't the `force` flag already do what the `skip_redownload_dataset` does?

JSabadin:
> Doesn't the `force` flag already do what the `skip_redownload_dataset` does?

They serve different purposes:

- `force` always triggers a fresh download, ignoring what’s on disk.
- `skip_redownload_dataset` simply prevents downloads if data already exists locally.

Probably a string parameter like `update_mode` could encapsulate all scenarios better:

- "always" = force a fresh download
- "if_empty" = only download if there’s no local data
- "never" = skip the download entirely

kozlov721:
> Probably a string parameter like `update_mode` could encapsulate all scenarios better:
> 
> - "always" = force a fresh download
> - "if_empty" = only download if there’s no local data
> - "never" = skip the download entirely

Merging those two parameters to one would be better. Imho it's not very clear what's the difference between `force=False` and `skip_redownloading_dataset=True`.

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -84,9 +85,9 @@ def __init__(
         @type width: Optional[int]
         @param width: The width of the output images. Defaults to
             C{None}.
-        @type force_resync: bool
-        @param force_resync: Flag to force resync from cloud. Defaults
-            to C{False}.
+        @param update_mode: Enum that determines the sync mode:
```
klemen1999, line 88: missing "@type update_mode ..."

klemen1999 (approved):

PR #222, opened 2025-01-14
Default empty task name

kozlov721:
Changed the default `task_name` from `"detection"` to an empty string. `"detection"` can be confusing for example for semantic segmentation tasks and there's not a reason to have any task name for datasets with only one task.


klemen1999 (approved):
LGTM, just needs fixing of tests

PR #224, opened 2025-01-15
Fixes for nested labels

kozlov721:
- Changed return type of `BaseDataset.get_tasks` to a dictionary mapping task names to task types (before it was just a list of combined task names)
- Changed return type of `BaseDataset.get_classes` from a tuple of all classes and a dictionary mapping task names to classes to just the later. The list of all classes doesn't make semantic sense with the new tasks, and also it was rarely even used anywhere
- Removed the `created_at` timestamp from parquet files - wasn't used anywhere
- Fixed `task_name` and `task_type` values for annotations with nested sub-detections
  - Names of the sub-detections are now part of the `task_name` instead of `task_type`
  - `{"car/driver": "boundingbox"}` instead of `{"car": "driver/boundingbox"}`
- Added additional overloads to methods loading parquet files 
- Simplified some tests

### DIFF
--- a/tests/test_data/test_dataset.py
+++ b/tests/test_data/test_dataset.py
@@ -2,10 +2,10 @@ from collections import defaultdict
 from pathlib import Path
 from typing import Any, Dict, List, Set
 
-import cv2
 import numpy as np
 import pytest
 from pytest_subtests.plugin import SubTests
+from utils import create_image
 
 from luxonis_ml.data import (
     BucketStorage,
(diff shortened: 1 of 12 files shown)


tests/test_data/test_dataset.py
```diff
@@ -540,22 +515,14 @@ def generator1():
     assert df_cloned.equals(df_original)
 
 
-@pytest.mark.dependency(
```
klemen1999, line 543: I believe this needs to stay in otherwise tests run in parallel and it can happen that GCS and Local tests both run at the same time on dataset with same name which corrupts the files. CC: @JSabadin to confirm

klemen1999 (approved):
LGTM, left one comment regarding tests

tests/test_data/test_dataset.py
kozlov721: That's no longer required if the `dataset_name` contains the bucket storage in it.

PR #228, opened 2025-01-17
Fix empty labels

kozlov721:
Added argument `exclude_empty_labels` to `LuxonisLoader`.
- If set to `True`, the output labels dictionary will not contain all the labels.
  - The dataset has a single task `"animals"` with two task types: `"boundingbox"` and `"segmentation"` 
  - One example doesn't have any animal in it $\rightarrow$ no boundingbox
  - The output label dictionary will have a single key `"animals/segmentation"`
- If set to `False` (default), the missing labels will be filled with empty arrays
  - Using the above example, the label dictionary will have two keys: `"animals/segmentation"` and `"animals/boundingbox"`
  - The shape of the empty label is compatible with a non-empty label
    - `[0, 5]` for the above example

### DIFF
--- a/luxonis_ml/data/augmentations/albumentations_engine.py
+++ b/luxonis_ml/data/augmentations/albumentations_engine.py
@@ -9,7 +9,7 @@ import numpy as np
 from typing_extensions import TypeAlias, override
 
 from luxonis_ml.data.utils.task_utils import get_task_name, task_is_metadata
-from luxonis_ml.typing import ConfigItem, LoaderOutput, TaskType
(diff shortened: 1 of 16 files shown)


klemen1999 (approved):
LGTM. Although I suppose this generally shouldn't be used with luxonis-train right? Because if used the `labels` dict will sometimes have e.g. "boundingbox" key and othertimes not.

kozlov721:
Yeah, it shouldn't be. I don't really know what the use case would be, but it might be useful eventually. We get it for free since the natural behavior is to drop empty labels and if we want to preserve them, we have to re-create them at the end.

PR #234, opened 2025-01-24
Metadata improvements

kozlov721:
- Added a `Category` type that can be used with the task metadata to signify the value is categorical as opposed to a general string
- Added new fields to the `LuxonisDataset.metadata`
  - `"categorical_encodings"` containing numerical encodings for categorical metadata tasks
  - `"metadata_types"` containing a mapping from metadata tasks to their python types
  - Added corresponding getters to `LuxonisDataset`
    - `get_categorical_encodings`
    - `get_metadata_types`  
- New parameter `keep_categorical_as_strings` in `LuxonisLoader`, default is `False`
  - If `False`, the loader output will contain the category ids instead of the string values
- Fixed a few small issues with metadata labels

This change allows for better flexibility in `luxonis-train` for models requiring metadata labels.

**Example:**

```python
from luxonis_ml.data import Category

def generator():
    yield {
        "task": "vehicles",
        ...,
        "metadata": {
            "color": Category("red")  # one of "red", "green", "blue"
            "license_plate": "xyz",
            "distance": 12.4,
            "id": 4,
        }
    }

dataset = LuxonisDataset(...)
dataset.add(generator)
assert dataset.get_categorical_encodings() == {
    "vehicles/metadata/color": {"red": 0, "green": 1, "blue": 2}}
}
assert dataset.get_metadata_types() == {
    "vehicles/metadata/color": "Category",
    "vehicles/metadata/license_plate": "str",
    "vehicles/metadata/distance": "float",
    "vehicles/metadata/id": "int",
}
```


luxonis_ml/data/datasets/annotation.py
```diff
@@ -25,11 +32,19 @@
 1]."""
 
 
+class Category(str):
```
kozlov721, line 35: Is better `Category` or `Categorical`?

klemen1999: I think Category is fine.

PR #238, opened 2025-01-29
fix mosaic4

JSabadin:
# Image Processing Improvements

**Key Fixes:**
- 🚫 **No more zoom effect** when loading images of different sizes than the original.
- 👁️ **Keypoints now auto-hide** when outside image bounds.

## Performance (1k images ×10 runs)

| Scenario               | Before  | After   | 
|------------------------|---------|---------|
| Load 128 (disk 256)    | 12.23s  | 12.33s  |
| Load 384 (disk 256)    | 12.54s  | 13.69s  |
| Matching sizes         | No change | 
*Note: Upsampling (slowest case) is rarely needed in practice*

## Visual Comparison (~640x460 → 384x384)

| Before                     | After                             |
|-----------------------|-------------------------|
|  | |

### DIFF
--- a/luxonis_ml/data/augmentations/custom/mosaic.py
+++ b/luxonis_ml/data/augmentations/custom/mosaic.py
@@ -496,11 +496,11 @@ def apply_mosaic4_to_bboxes(
         shift_x = out_width
         shift_y = out_height
 
-    bbox[:, [0, 2]] += shift_x - x_crop
-    bbox[:, [1, 3]] += shift_y - y_crop
+    bbox[:, 0] += shift_x - x_crop
+    bbox[:, 2] += shift_x - x_crop
 
-    bbox[:, [0, 2]] = np.clip(bbox[:, [0, 2]], 0, out_width)
-    bbox[:, [1, 3]] = np.clip(bbox[:, [1, 3]], 0, out_height)
+    bbox[:, 1] += shift_y - y_crop
+    bbox[:, 3] += shift_y - y_crop
 
     bbox = normalize_bboxes(bbox, (out_height, out_width))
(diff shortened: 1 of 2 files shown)


luxonis_ml/data/augmentations/custom/mosaic.py
```diff
@@ -496,11 +496,11 @@ def apply_mosaic4_to_bboxes(
...
 
-    bbox[:, [0, 2]] += shift_x - x_crop
-    bbox[:, [1, 3]] += shift_y - y_crop
+    bbox[:, 0] += shift_x - x_crop
+    bbox[:, 2] += shift_x - x_crop
 
-    bbox[:, [0, 2]] = np.clip(bbox[:, [0, 2]], 0, out_width)
-    bbox[:, [1, 3]] = np.clip(bbox[:, [1, 3]], 0, out_height)
+    bbox[:, 1] += shift_y - y_crop
+    bbox[:, 3] += shift_y - y_crop
```
klemen1999, line 503: Any reason why we don't perform clipping anymore?

JSabadin: Yeah, great question! The short answer is that Albumentations handles all the label clipping, selection of the best labels, and discarding labels that are out of bounds. Additionally, for bounding boxes that are entirely out of bounds, this code assigns them a bounding box with an area of 0. 

However, the important part is that we should set the visibility of keypoints to 0 if they are outside the bounds. This was not previously implemented, and as you can see in the image (before), such keypoints were stacked on the edges.

klemen1999 (approved):

PR #243, opened 2025-02-06
fix kpts and bbox edge cases

JSabadin:
## Purpose

We've introduced a bounding-box area threshold (`area_threshold = 0.0001`) to remove very small/invalid bounding boxes and their corresponding keypoints. We also added edge-case handling for keypoints.

## Specification

- **Filtering**: `postprocess_bboxes` now excludes boxes with `width * height < 0.0001`.
- **Keypoint Indexing**: `postprocess_keypoints` discards keypoints for excluded boxes and clamps out-of-bounds coordinates setting the visibility to 0.
## Testing & Validation

Before:
- Keypoints appear at (0,0) with v > 0.
- Minimal-area bounding boxes remain
- Out-of-range keypoints have visibility=2
Before

After 
After

### DIFF
--- a/luxonis_ml/data/augmentations/utils.py
+++ b/luxonis_ml/data/augmentations/utils.py
@@ -39,24 +39,15 @@ def postprocess_mask(mask: np.ndarray) -> np.ndarray:
 
 
 def postprocess_bboxes(bboxes: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
-    area_threshold = 0.0004  # 0.02 * 0.02 Small area threshold to remove invalid bboxes and respective keypoints.
     if bboxes.size == 0:
-        return np.zeros((0, 5)), np.zeros((0,), dtype=np.uint8)
+        return np.zeros((0, 6)), np.zeros((0, 1), dtype=np.uint8)
+
     ordering = bboxes[:, -1]
-    raw_bboxes = bboxes[:, :-1]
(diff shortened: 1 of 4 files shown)


luxonis_ml/data/augmentations/utils.py
```diff
@@ -39,15 +39,24 @@ def postprocess_mask(mask: np.ndarray) -> np.ndarray:
 
 
 def postprocess_bboxes(bboxes: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    area_threshold = 0.0001  # 0.01 * 0.01 Small area threshold to remove invalid bboxes and respective keypoints.
```
kozlov721, line 42: There's already an argument in `__init__` of `AblumentationsEngine` called `min_bbox_visibility` that should do the same thing. It's set to `0` by default, would setting it to `1e-4` work the same as this change?

kozlov721 (commented):
Left a comment regarding the `area_threshold`. Also, could you please add a few test cases for the specific examples that used to fail before? Otherwise LGTM

luxonis_ml/data/augmentations/utils.py
JSabadin: @kozlov721 @klemen1999 
`min_visibility `sets the **minimum fraction of the original bounding box** that must remain visible after augmentation. If a transformed box’s visible area falls below this threshold, it will be dropped from the output.

klemen1999 (approved):

PR #245, opened 2025-02-07
Sort classes alphabetically

JSabadin:
## Purpose

Helps to keep consistent order of class names.

## Specification

- Changed signature of `BaseDataset.get_classes` and `set_classes` to accept and return a mapping from class names to IDs instead of a list of class names
  - `set_classes` accepts both a list of class names and the mapping
    - If list is provided, the class IDs are assigned alphabetically (except for the `"background"` class)
    - If mapping is provided, it is used as is
      - In case we would want to explicitly set a specific class IDs
- `LuxonisDataset._metadata` changed from a dictionary to a `pydantic` model
  - The same for `LuxonisSource` and `LuxonisComponent`
- Cleaned up the `luxonis_dataset.py` a bit
  - Moved the `Metadata` class and migration code to separate files

Change of the `BaseDataset.get_classes` and `BaseDataset.set_classes` is a breaking change

## Testing & Validation
Added new test file `test_classes_ordering.py` that tests the ordering in both `LuxonisDataset` and `LuxonisLoader`

### DIFF
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -655,9 +655,9 @@ class LuxonisDataset(BaseDataset):
             )
             metadata_json = json.loads(path.read_text())
             version = Version.parse(metadata_json.get("ldf_version", "1.0.0"))
-            if version == LDF_VERSION:
-                return Metadata(**metadata_json)
-            return self._migrate_metadata(metadata_json)
+            if version != LDF_VERSION:  # pragma: no cover
+                return self._migrate_metadata(metadata_json)
+            return Metadata(**metadata_json)
         else:
             return Metadata(
                 source=LuxonisSource(),
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -1030,20 +1030,9 @@ class LuxonisDataset(BaseDataset):
                 logger.info(
                     f"Detected new classes for task {task}: {new_classes}"
                 )
-                all_class_names = classes | old_classes
-                has_background = "background" in all_class_names
-                all_class_names -= {"background"}
-                all_class_names = sorted(all_class_names)
-                if has_background:
-                    all_class_names.insert(0, "background")
 
-                self.set_classes(
-                    {
-                        class_name: i
-                        for i, class_name in enumerate(all_class_names)
-                    },
-                    task=task,
-                )
+                self.set_classes(list(classes | old_classes), task=task)
+
         for task, num_kpts in num_kpts_per_task.items():
             self.set_skeletons(
                 labels=[str(i) for i in range(num_kpts)],
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -489,7 +489,7 @@ class LuxonisDataset(BaseDataset):
 
         if not attempt_migration or self.version == LDF_VERSION or df is None:
             return df
-        return migrate_dataframe(df)
+        return migrate_dataframe(df)  # pragma: no cover
 
     @overload
     def _get_file_index(
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -18,9 +18,7 @@ from typing import (
     Sequence,
     Set,
     Tuple,
-    TypedDict,
     Union,
-    cast,
     overload,
 )
 
@@ -29,7 +27,6 @@ import polars as pl
 import pyarrow.parquet as pq
 from filelock import FileLock
 from loguru import logger
-from ordered_set import OrderedSet
 from semver.version import Version
 from typing_extensions import Self, override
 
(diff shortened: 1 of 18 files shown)


tests/test_data/test_loader.py
```diff
@@ -308,3 +319,175 @@ def generator():
...
+    dataset = LuxonisDataset(
+        "test_classes_order",
+        delete_existing=True,
+        delete_remote=True,
+        bucket_storage=BucketStorage.LOCAL,
+    )
+    dataset.add(generator())
+    dataset.make_splits(ratios=(1, 0, 0))
+    classes = dataset.get_classes()[""]
+    expected_classes = (
```
klemen1999, line 351: Let's use the `sorted(classes, key=lambda x: (0 if x == "background" else 1, x))`, seems cleaner (also anywhere else this applies)

tests/test_data/test_loader.py
```diff
@@ -308,3 +319,175 @@ def generator():
...
+    assert (
+        classes == expected_classes
+    ), f"Expected {expected_classes}, but got {classes}"
+
+    loader = LuxonisLoader(dataset, view="train", height=img_h, width=img_w)
+    for batch in loader:
+        classification = batch[1]["/classification"]
+        assert len(classification) == len(classes_list)
+
+    mapping = sorted(
```
klemen1999, line 445: Here sorting shouldn't be needed since class_mappings is constructed by iteratng self.classes and self.classes should be already sorted correctly ("background" first and then the rest sorted)

tests/test_data/test_loader.py
```diff
@@ -308,3 +319,175 @@ def generator():
...
+        delete_remote=True,
+        bucket_storage=BucketStorage.LOCAL,
+    )
+    dataset.add(generator())
+    dataset.make_splits(ratios=(1, 0, 0))
+    classes = dataset.get_classes()[""]
+    expected_classes = (
+        ["background"] if "background" in classes else []
+    ) + sorted([c for c in classes_list if c != "background"])
+    assert (
```
klemen1999, line 354: These checks on the dataset.get_classes() should be also done at the end, after LuxonisLoader init because LuxonisLoader can change their state (same for other test functions)

klemen1999 (commented):
[Here](https://github.com/luxonis/luxonis-ml/blob/05b36c38a6eaf7c56063e95937310ba20924e566/luxonis_ml/data/loaders/luxonis_loader.py#L191) the background class should be added at index 0 not at the end otherwise dataset.get_classes() won't have "background" at first index.

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -169,8 +169,7 @@ def __init__(
         for task, seg_masks in task_type_iterator(test_labels, "segmentation"):
             task_name = get_task_name(task)
             if seg_masks.shape[0] > 1 and (
-                "background" not in self.class_mappings
```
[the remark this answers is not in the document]
JSabadin: This expression will always evaluate to `True `because `self.class_mappings` is a dictionary of dictionaries. For example, if `self.class_mappings` is defined as:
```
self.class_mappings = {
    "": {'background': 0, 'cat': 1, 'dog': 2}
}
```
then the outer dictionary has only one key—the empty string (""). Since 'background' is a key in the inner dictionary and not in the outer one, the expression "background" not in self.class_mappings returns True.


Better way would be:
`            if seg_masks.shape[0] > 1 and (
                self.class_mappings[task_name].get("background", -1) != 0
            ):`

tests/test_data/test_loader.py
JSabadin: Sorting is needed because it will happen 
`{'airplane': 1, 'cat': 2, 'dog': 3, 'background': 0}` -> list -> `['airplane', 'cat', 'dog', 'background']` instead of `['background', 'airplane', 'cat', 'dog']`

luxonis_ml/data/__main__.py
```diff
@@ -44,14 +44,19 @@ def check_exists(name: str):
 
 def get_dataset_info(dataset: LuxonisDataset) -> Tuple[Set[str], List[str]]:
     all_classes = {
-        c for classes in dataset.get_classes().values() for c in classes
+        class_name
```
klemen1999, line 47: For `info` it might be actually better to just print the whole mapping since we have it

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -729,8 +730,9 @@ def _migrate_metadata(
     ) -> Metadata:  # pragma: no cover
         old_classes = metadata["classes"]
```
klemen1999, line 731: Won't this crash now because hte Metadata doesn't have "classes" key anymore?

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -729,8 +730,9 @@ def _migrate_metadata(
...
         if set(old_classes.keys()) <= LDF_1_0_0_TASKS:
-            metadata["classes"] = {
-                "detection": next(iter(old_classes.values()))
+            classes = list(next(iter(old_classes.values())))
```
klemen1999, line 733: Should old_classes be sorted before we create class_mappings? To ensure that values in the mapping are correct i.e. "background" (if present) has 0, "animal" has 1, "bed" has 3 etc.
Or even better we should probably use .set_classes() which already handles this

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -778,7 +786,11 @@ def set_classes(
             tasks = [task]
 
         for task in tasks:
-            self._metadata["classes"][task] = classes
+            sorted_classes = (
```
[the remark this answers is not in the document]
klemen1999: same comment in case it was missed

tests/test_data/test_loader.py
```diff
@@ -308,3 +319,160 @@ def generator():
...
+    loader = LuxonisLoader(dataset, view="train", height=img_h, width=img_w)
+    for batch in loader:
+        classification = batch[1]["/classification"]
+        assert len(classification) == len(classes_list) + 1
+        assert not np.array_equal(
+            classification, [1.0] + [0.0] * len(classes_list)
+        )
+    sorted_classes = ["background"] + sorted_classes
+    for i in range(len(classes_list)):
+        assert loader.class_mappings[""][sorted_classes[i]] == i
```
klemen1999, line 397: IMO we should also add `dataset.get_classes() `tests here after the loader init. It should be same as `loader.class_mappings` but it would be good to specifically check (same comment for other tests)

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -357,15 +302,15 @@ def clone(
         )
 
         new_dataset._init_paths()
-        new_dataset._metadata = defaultdict(dict, self._get_metadata())
+        new_dataset._metadata = self._get_metadata()
 
-        new_dataset._metadata["original_dataset"] = self.dataset_name
+        new_dataset._metadata.parent_dataset = self.dataset_name
```
kozlov721, line 307: Do we need to store this info?

klemen1999: I don't think it's critical. If it adds to complexity then we can skip it imo

kozlov721: It doesn't really add any complexity; I was just wondering what the use case for it would be. It can stay in

klemen1999 (approved):
LGTM

PR #247, opened 2025-02-14
Removed deprecated parameters from `make_splits`

kozlov721:
## Purpose

`make_splits` carried three ways to say the same thing: `splits`, plus the deprecated
`ratios` and `definitions` keywords. This PR removes the two deprecated ones. One
argument now has one meaning.

The cleanup exposed several real defects in the split logic. This PR fixes them too.
Each fix has a regression test.

## Specification

**1. The API**

`BaseDataset.make_splits` and `LuxonisDataset.make_splits` lose `ratios` and
`definitions`. `splits` is the only entry point. It still accepts every shape:

| `splits` | Meaning |
| --- | --- |
| `{"train": 0.8, "val": 0.1, "test": 0.1}` | ratios by split name |
| `(0.8, 0.1, 0.1)` | ratios for train, val, and test |
| `{"train": ["a.jpg", …]}` | explicit filepaths |
| `None` | the default 80/10/10 |

A ratio is any number from 0 to 1, and the ratios sum to 1. `1`, `1.0`, and `True` all
mean the same thing, in a mapping and in a tuple. `make_splits` reads a ratio and never
a count, so there is nothing to disambiguate. `BaseParser` resolves counts one layer up:
it converts a count mapping to explicit filepath lists before it calls the method.

`BaseDataset.make_splits` widens its type from `dict` to `Mapping`.
`LuxonisDataset.make_splits` already used `Mapping`. A tuple of filepaths now works.
`main` accepted one only when the first value of the mapping was a `list`: a later tuple
raised `TypeError("Must provide splits as a list of filepaths")`, and a mapping of
tuples fell through to the default 80/10/10 in silence.

`make_splits` now holds the orchestration only. `_resolve_splits` reads the argument and
validates the ratios. `_split_by_ratio` and `_split_by_definition` hold one strategy
each. The method loads the old splits, dispatches, merges, and writes.

**2. Defects fixed along the way**

| Fix | Was |
| --- | --- |
| Divide the splits by the largest remainder method | `ceil` gave each split more than its share in turn, so the last split absorbed all the error. With the default 80/10/10 ratios, the test split was empty for 9 groups or fewer, and again for 11 to 14 groups. A 7-group dataset gave 6/1/0; it now gives 5/1/1. Above 14 groups the test split survived, but train still rounded up whenever its exact share was fractional: 17 groups gave train 14 against a share of 13.6. An empty split with a non-zero ratio now warns. |
| Write `splits.json` atomically | The write went straight to `splits.json`. A failure in the middle left a truncated file, so the old splits were gone. |
| Honour `replace_old_splits` for explicit definitions | The flag did nothing for definitions. The method always added to the old splits. |
| Skip a group that already belongs to a split | The definitions branch deduplicated inside one split of one call, and nothing more. It never read the existing splits, so a second call with the same definitions wrote every group again. A 6-group dataset went from 4/2 to 8/4. |
| Keep one file out of two splits | `{"train": [a, b], "val": [a]}` put `a` in train **and** in val, so a sample leaked between the two. The first split that lists a file now keeps it. |
| Read every value of a splits mapping, not only the first | The method read only the **first** value. A value that was neither a `float` nor a `list` left both branches unset, so the method silently used the default 80/10/10 and discarded the argument. On 20 groups, `{"train": 8, "val": 1, "test": 1}` gave 16/2/2 rather than an error; `{"train": 1}` and `{"train": "invalid"}` did the same. The result also changed with the key order: `{"train": 0.5, "val": 0.5, "test": 0}` gave 10/10/0, but the same values as `{"test": 0, "train": 0.5, "val": 0.5}` gave 16/2/2. Both now give 10/10/0, and a count mapping now fails with `Ratios must be between 0.0 and 1.0`. |
| Warn and skip a non-filepath definition element | An `int` in a filepath list reached `Path()`. It raised `expected str, bytes or os.PathLike object, not int`, far from the caller. |
| Add a range check for the ratios | The method only checked the sum. `{"train": 1.5, "val": -0.5}` sums to 1.0, so it passed. It then gave train every group and val none, in silence. |
| `remove_duplicates` skips a no-op rewrite | It rewrote every parquet file. It also re-uploaded the whole `annotations` directory, even when it removed nothing. |

**3. One open question**

`main` deduplicated inside one split of one call, and nothing more. It never read the
existing splits, so a second call with the same filepaths added every group id again. A
6-group dataset went from 4/2 to 8/4. Its docstring promised more than the code did:
"duplicate group IDs will be filtered out".

The method now skips a file that already sits in a split. If nothing is new, it warns
and leaves the splits as they are.

The alternative is to raise, which is what the ratio path does when it finds nothing
new. Either way `replace_old_splits=True` stays the only way to move a file between
splits, so the choice is only whether the caller hears about it.

The two paths disagree today, so this needs a decision either way.

Breaking, and deliberately so:

- **`ratios` and `definitions` are gone.** Callers pass `splits` instead. This is the
  point of the PR.
- **Split sizes change.** The largest remainder method changes the result for many group
  counts, not only small ones. With the default ratios, 17 groups gave 14/2/1 and now
  give 13/2/2. Between 15 and 60 groups, 31 of the 46 counts change. A 4-group dataset
  gave 4/0/0 and now gives 3/1/0.
- **`replace_old_splits=True` now always discards the old splits.** The flag did nothing
  for explicit definitions. With ratios it only took effect when no unassigned file
  remained.
- **Explicit definitions skip a group that already belongs to a split.** A repeated call
  with the same definitions is now a no-op. It previously added the group a second time.

Confined to `make_splits` and its helpers. `LuxonisLoader`, the parsers, and the dataset
format stay untouched.

## Deployment Plan

None / not applicable. This is a library change with no rollout step.

## Testing & Validation

The split tests live in `tests/test_data/test_splits.py`.

Every fix has a regression test. Each one also ran against the reverted fix, so it
fails without it:

| Test | Fails without the fix with |
| --- | --- |
| `test_split_sizes` | totals that do not sum to the group count |
| `test_small_dataset_keeps_every_split` | `{'train': 6, 'val': 1, 'test': 0}` for 7 groups |
| `test_make_splits_is_atomic` | `json.decoder.JSONDecodeError` on a truncated `splits.json` |
| `test_definitions_skip_a_non_filepath` | `TypeError: expected str, bytes or os.PathLike object, not int` |
| `test_remove_duplicates_skips_a_clean_dataset` | a full 3-row rewrite on a dataset with no duplicates |

`make_splits` runs against a real dataset for every shape the signature accepts: `None`,
a tuple of ratios, a mapping of ratios, and a mapping of filepath lists. The filepaths
go in as `str` and as `Path`, inside a `list` and inside a `tuple`. Custom split names
and `replace_old_splits=True` each get a case. One test pins the leak: no group may land
in two splits.

`_split_sizes` and `_resolve_splits` are pure, so hypothesis covers them. Every group
lands in exactly one split, the sizes sum to the group count, a zero ratio takes
nothing, and `1`, `1.0`, and `True` resolve to one ratio.

`ruff`, `pyright --warnings`, and the `tests/test_data` suite are clean.

## AI Usage

Assisted-by: Claude Code:claude-opus-5

## Summary by CodeRabbit

* **Improvements**
  * Simplified dataset split configuration with a unified interface supporting ratios, file mappings, or default splits.
  * Added replacement of existing splits while preventing duplicate assignments.
  * Improved validation, ratio handling, and error reporting for invalid definitions and unavailable data.
  * Preserved existing assignments when explicit file-based splits add no new files.
  * Improved reliability by preventing unnecessary updates and protecting split data during write failures.

* **Documentation**
  * Updated split-generation guidance and replacement-option terminology.

* **Tests**
  * Expanded coverage for validation, replacement behavior, duplicate definitions, and edge cases.


luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -1003,92 +1002,111 @@ def get_splits(self) -> Optional[Dict[str, List[str]]]:
...
+            new_splits = self._make_ratio_splits(
+                splits,  # type: ignore
+                old_splits,
+                replace_old_splits,
+            )
+        else:
+            new_splits = self._make_definition_splits(splits)  # type: ignore
+
+        for split, uuids in new_splits.items():
+            old_splits[split].extend(uuids)
```
klemen1999, line 1045: We should only extend if `replace_old_splits=False` no? Otherwise if split_type="files" wouldn't we in this case now get one image in two times in the splits (once from older split and once from newer one)?

luxonis_ml/data/datasets/luxonis_dataset.py
```diff
@@ -1494,170 +1487,84 @@ def make_splits(
             sources (``"train_real"``, ``"train_synth"``) or for
             creating fully custom splits.
 
+        Note:
+            The method skips each file that is already in a split. Thus
```
klemen1999: I'd remove "Thus you can give the same filepath definitions more than one time." sentence as it doesn't make sense. You can technically do this but you don't want to so we shouldn't even note it.

klemen1999 (approved):
Generally LGTM

PR #256, opened 2025-02-26
fix examples

JSabadin:
## Purpose

Aligning the examples with the current luxonis-ml version.


klemen1999:
Could you also update the README in `luxonis_ml/data`? We should make the "adding data" code block more informative (can be something similar to example [here](https://github.com/luxonis/luxonis-ml/pull/211)). Also the `Annotation format` needs to be updated (the `type` is not used anymore, also other parts). Also the `Usage with LuxonisLoader` needs updating. 
Overall let's make sure all the docs are up to date

klemen1999 (approved):

PR #265, opened 2025-03-04
Fix Random Seed Handling and Add Augmentation Tests

JSabadin:
## Purpose

This update fixes albumentations to 1.4.18 for Python 3.8/3.10 compatibility and external seed control, adds extra tests, and ensures deterministic dataset creation.

## Specification

None / not applicable

- We hardcode albumentations to 1.4.18 so that it is compatible with Python 3.8 and 3.10, and the random seed can be set outside of albumentations.
- Additional tests have been added.
- We have made the dataset creation deterministic.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


luxonis_ml/data/augmentations/albumentations_engine.py
```diff
@@ -384,7 +386,7 @@ def get_params(is_custom: bool = False) -> Dict[str, Any]:
                 A.Compose(spatial_transforms, **get_params())
             )
             self.pixel_transform = wrap_transform(
-                A.Compose(pixel_transforms), is_pixel=True
+                A.Compose(pixel_transforms, seed=seed), is_pixel=True
```
klemen1999, line 389: Don't we need to specify `seed` also for other `Compose`s (spatial and custom, maybe even resize to be safe)?

luxonis_ml/data/augmentations/albumentations_engine.py
```diff
@@ -249,6 +249,7 @@ def __init__(
         keep_aspect_ratio: bool = True,
         is_validation_pipeline: bool = False,
         min_bbox_visibility: float = 0.0,
+        seed: int | None = None,
```
klemen1999, line 252:
```suggestion
        seed: Optional[int] = None,
```
because of python3.8

klemen1999, line 252: I think we need to add the `seed` also in base_engine.py init

luxonis_ml/data/augmentations/albumentations_engine.py
JSabadin: We do it in the `get_params`

luxonis_ml/data/augmentations/albumentations_engine.py
```diff
@@ -384,7 +395,7 @@ def get_params(is_custom: bool = False) -> Dict[str, Any]:
                 A.Compose(spatial_transforms, **get_params())
             )
             self.pixel_transform = wrap_transform(
-                A.Compose(pixel_transforms), is_pixel=True
+                A.Compose(pixel_transforms, seed=seed), is_pixel=True
```
klemen1999, line 398: Is this a real type check error? If it ins't we can explicitly ignore so it doesn't pop up

klemen1999: `seed` argument is present only from version `1.4.21` up. I updated the requirements accordingly

klemen1999:
Actually the issue is a bit deeper: `Seed` is introduced in the `albumentations>=1.4.21` ([here](https://github.com/albumentations-team/albumentations/releases?page=1#:~:text=Random%20state,global%20random%20state)) but the problem is that for py3.8 the newest supported version is `1.4.18`. So this is not only windows thing but a version thing.
We can check what is the version and at runtime decide to set seed or not but overall IMO it's better to actually hardcode to only one specific albumentations version (in this case it would have to be 1.4.18 because of py3.8) because it seems they change some nuances quite often between versions and we rely on certain flow for our engine to work correctly (e.g. Mosaic and filtering of boxes, keypoints that fall outside)

PR #280, opened 2025-03-11
Feature: safe YAML

kozlov721:
## Purpose

Makes it possible to load back saved config using safe loader.

## Specification

Changed `yaml.damp` to `yaml.safe_dump`

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


klemen1999 (approved):
LGTM but to be sure can we add a simple test for it that goes over the configs yamls, loads them in the config, saves it back to yaml and then tries to load them back?

PR #282, opened 2025-03-12
Fix Dataset Corruption from Missing Class Info in Last Yield

JSabadin:
## Purpose

Fixed an issue where, if we yielded the annotation as the last yield without the class info but with the instance_id, the dataset would become corrupted. See the image under testing & validation.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Previously, if the keypoints were yielded at the end with missing class information but with the `instance_id` (so they should be linked with the bounding boxes), the dataset would be corrupted:

A test was added to check for this issue, where the classes would appear as: `Classes: {'': {}}`.

### DIFF
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -923,7 +923,7 @@ class LuxonisDataset(BaseDataset):
                     def update_state(task_name: str, ann: Detection) -> None:
                         if ann.class_name is not None:
                             classes_per_task[task_name].add(ann.class_name)
-                        else:
+                        elif not classes_per_task[task_name]:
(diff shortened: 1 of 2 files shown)


kozlov721 (approved):
LGTM, though in the future we probably shouldn't allow generator like this (no `"class_name"`). Currently the `class_name` is optional because you technically can have a dataset where it doesn't make sense, but I think in these rare situations it's better to force filling in the `class_name` anyways as opposed to dealing with possible `None` in all the other cases.

klemen1999:
> though in the future we probably shouldn't allow generator like this (no "class_name").
Agree with this, I'm fine with making it a required field always

PR #284, opened 2025-03-12
Fix failing keypoints loader

JSabadin:
## Purpose

If the dataset only contains keypoints, the loader would fail. This PR fixes that issue and adds a test for it.
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


tests/test_data/test_dataset.py
```diff
@@ -700,3 +700,28 @@ def generator() -> DatasetIterator:
...
+                "file": img,
+                "annotation": {
+                    "class": "person",
+                    "keypoints": {"keypoints": keypoints},
+                    "instance_id": instance_id,
+                },
+            }
+
+    dataset = create_dataset(dataset_name, generator())
+    loader = LuxonisLoader(dataset)
```
kozlov721, line 725: We should also test with augmentations

kozlov721 (approved):

PR #286, opened 2025-03-13
Feat/docs

JSabadin:
## Purpose

The purpose is to improve the documentation for first-time users.

## Specification

- Updates to the original data documentation.
- Added README files for the `LuxonisParser`, `LuxonisLoader`, and `LuxonisDataset` parameters, and linked them in the main data README to provide users with more information on what to configure.
- Made the generator example more informative by including `task_name` and explaining its purpose.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_ml/data/__main__.py
+++ b/luxonis_ml/data/__main__.py
@@ -85,7 +85,9 @@ def print_info(dataset: LuxonisDataset) -> None:
         task_table.add_column(
             "Task Name", header_style="magenta i", max_width=30
         )
-    task_table.add_column("Task Types", header_style="magenta i", max_width=50)
+    task_table.add_column(
+        "Annotation Types", header_style="magenta i", max_width=50
+    )
     for task_name, task_types in tasks.items():
         task_types.sort()
         if not task_name:
--- a/luxonis_ml/data/README.md
+++ b/luxonis_ml/data/README.md
@@ -50,7 +50,7 @@ Each of these steps will be explained in more detail in the following examples.
 
 We will be using our toy dataset `parking_lot` in all examples. The dataset consists of images of cars and motorcycles in a parking lot. Each image has a corresponding annotation in the form of a bounding box, keypoints and several segmentation masks.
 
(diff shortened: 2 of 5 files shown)


luxonis_ml/data/__main__.py
```diff
@@ -85,7 +85,9 @@ def print_info(dataset: LuxonisDataset) -> None:
         task_table.add_column(
             "Task Name", header_style="magenta i", max_width=30
         )
-    task_table.add_column("Task Types", header_style="magenta i", max_width=50)
+    task_table.add_column(
+        "Annotation Types", header_style="magenta i", max_width=50
```
klemen1999, line 89: Generally renaming task type into annotation type does make sense but I don't want to add more confusion around this since we also use `task` keyword on luxonis-train - it doesn't mean completenly the same but general user I think will understand it as the same (e.g. I want to train bbox detection and I have bbox task type in dataset and bbox task set in the head). 
CC: @kozlov721 also for thoughts on this if worth to change

klemen1999 (commented):
LGTM, just left one comment on "task type -> annotation type" rename

luxonis_ml/data/__main__.py
kozlov721: IMHO the best solution would be to rename the `task` here in `luxonis-ml` to something completely different, like `label_group` or `annotation_group`, so there's no confusion about what `task` is supposed to mean. 

The current naming convention in `luxonis-ml` is
```
       task
        |
/-----------------\
vehicles/boundingbox
\------/ \---------/
   |          |
task_name    task_type 
```

The `task_name` would be the `label_group` and `task_type` can be `label_type`. And the full thing can be called `label_name`.

(and while we're at it we should decide whether we're saying `annotations` or `labels` and keep it consistent everywhere)

kozlov721 (approved):

klemen1999 (approved):

PR #294, opened 2025-03-28
Add Keypoint Labels Plotting Next to Keypoints

JSabadin:
## Purpose

Added plotting of keypoint labels next to the keypoints.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


luxonis_ml/data/utils/visualizations.py
```diff
@@ -458,13 +458,31 @@ def create_mask(
...
                 draw_function(
                     curr_image,
-                    (int(k[0] * w), int(k[1] * h)),
+                    point,
                     size,
                     color=color,
                     thickness=2,
                 )
+
+                font_scale = 0.4
```
klemen1999, line 470: Can we make the font_scale depend on the min(h,w)? Some factor of this value so on smaller images this is small but on bigger images this text is appropriatly scaled so it can be actually seen?

klemen1999 (approved):
LGTM

PR #298, opened 2025-04-07
Dataset Push/Pull/Clone/Merge

JSabadin:
## Purpose

- Removed redundant `file_index` (all necessary information is already stored in the Parquet files).

- Fixed an issue where a dataset without `metadata.json` was considered non-existing. Now, a dataset is recognized as existing if it contains a `metadata` folder.

- Updated the `add` function: if called on an existing dataset and the user provides the same paths as those in the Parquet file, the data will be overwritten. However, duplicates within the current batch are not checked - but  this is a user error that can be detected using the `dataset health` command. Additionally, the `add` function will warn the user at the end with`warn_on_duplicates` .

- When using `UpdateMode.MISSING`, only missing media file paths are synced. Annotations and metadata are always synced (pushed or pulled, depending on the operation).

- Added CLI commands for push, pull, clone, and merge, along with tests for these dataset operations.

- Enabled pushing a dataset from local storage to the cloud, deleting it locally, pulling it back from the cloud, modifying it, and pushing it again... Merging with other datasets is also supported (datasests should be on the same bucket to merge).

- Removed non-intuitive logic for dataset deletion. Since a dataset can exist locally or in the cloud, deletion should explicitly target either or both.

- Fixed a bug in the print command trigered when one task had an empty string and the other did not.

- Added documentation on dataset structure and Parquet files. Clarified that when yielding data without providing `instance_id`s for keypoints and boxes, those annotations must be yielded together—otherwise, the loader will not behave as expected.

- Slightly improved tests and added a new test for push/pull functionality.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_ml/data/datasets/luxonis_dataset.py
+++ b/luxonis_ml/data/datasets/luxonis_dataset.py
@@ -834,7 +834,7 @@ class LuxonisDataset(BaseDataset):
                 local_paths=self.local_path / dir_name,
                 remote_dir=dir_name,
                 copy_contents=True,
-            )  # Not sure about that
+            )
 
         if update_mode == UpdateMode.ALL:
             logger.info("Force-pushing all media files...")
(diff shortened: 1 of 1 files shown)


kozlov721 (approved):
Looks pretty good to me, thanks!

klemen1999 (approved):

PR #302, opened 2025-04-17
Show bbox class labels and scale keypoints dynamically

JSabadin:
## Purpose

Display the bounding box class next to the bounding box. Use a dynamic scale for keypoints.

image

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


luxonis_ml/data/utils/visualizations.py
```diff
@@ -254,8 +254,6 @@ def create_text_image(
     """
     img = np.full((height, width, 3), resolve_color(bg_color), dtype=np.uint8)
 
-    font = cv2.FONT_HERSHEY_SIMPLEX
```
klemen1999, line 257: Isn't this function now missing `font` variable but `font` is used in the `text_size` call?

luxonis_ml/data/utils/visualizations.py
```diff
@@ -327,6 +325,96 @@ def concat_images(
...
+        x1, y1, x2, y2], where (x1, y1) is the top-left corner and (x2,
+        y2) is the bottom-right corner.
+    @type color: Tuple[int, int, int]
+    @param color: The color of the label.
+    @type font_scale: float
+    @param font_scale: The scale of the font.
+    """
+    text = class_name
+
+    text_size = cv2.getTextSize(text, font, font_scale, 1)[0]
```
klemen1999, line 355: Same thing: `font` is used but not defined anywhere in the function?

luxonis_ml/data/utils/visualizations.py
JSabadin: It's set at the beginning of the script as a global variable

luxonis_ml/data/utils/visualizations.py
JSabadin: Same answer as above.

luxonis_ml/data/utils/visualizations.py
klemen1999: Oh I see it now. In that case lets use all caps which is the usual in python for global variables

klemen1999 (approved):

PR #303, opened 2025-04-17
albumentations>=2.0.0

JSabadin:
## Purpose

Upgrade to Albumentations version 2.0.0 or higher.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_ml/data/augmentations/custom/letterbox_resize.py
+++ b/luxonis_ml/data/augmentations/custom/letterbox_resize.py
@@ -59,35 +59,31 @@ class LetterboxResize(A.DualTransform):
         return targets
 
     @override
-    def update_params(
-        self, params: Dict[str, Any], **kwargs
+    def get_params_dependent_on_data(
+        self, params: Dict[str, Any], data: Dict[str, Any]
     ) -> Dict[str, Any]:
         """Updates augmentation parameters with the necessary metadata.
 
         @param params: The existing augmentation parameters dictionary.
         @type params: Dict[str, Any]
-        @param kwargs: Additional keyword arguments to add the
-            parameters.
-        @type kwargs: Any
-        @return: Updated dictionary containing the merged parameters.
+        @param data: The data dictionary.
+        @type data: Dict[str, Any]
+        @return: Additional parameters for the augmentation.
         @rtype: Dict[str, Any]
         """
+        rows, cols, _ = params["shape"]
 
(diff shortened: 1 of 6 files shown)


kozlov721:
We should maybe wait a bit with merging since this PR effectively pushes `luxonis-ml` to Python 3.10. I'd say it would be better to first open and merge a PR that only does the migration and then merge this one so it's easier to rollback in case it's needed.

klemen1999:
> We should maybe wait a bit with merging since this PR effectively pushes luxonis-ml to Python 3.10

Yeah totally agree, this is just a "preparation PR". A separate "bump to py3.10" PR will get merged first

luxonis_ml/data/augmentations/custom/letterbox_resize.py
```diff
@@ -59,35 +59,31 @@ def targets(self) -> dict[str, Any]:
...
-        @param kwargs: Additional keyword arguments to add the
-            parameters.
-        @type kwargs: Any
-        @return: Updated dictionary containing the merged parameters.
+        @param data: The data dictionary.
+        @type data: Dict[str, Any]
+        @return: Additional parameters for the augmentation.
         @rtype: Dict[str, Any]
         """
+        cols, rows, _ = params["shape"]
```
klemen1999, line 74: Any particular reason we swapped what was cols and rows before? If "shape' is in HWC format then first component would be rows right, as in how many rows on y axis does the arra have?

luxonis_ml/data/requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy~=1.22
 opencv-python~=4.10
-albumentations>=1.4.16,<1.4.21
+albumentations>=2.0.0
```
klemen1999, line 3: Does it work with all versions from 2.0.0 to latest? Ww should run tests on all so we are sure because historically albumentations doesn't really announce breaking changes clearly

JSabadin: I’ve tested it locally on all of them, and it worked

luxonis_ml/data/augmentations/custom/letterbox_resize.py
JSabadin: The order is ‘H, W, C’, so height corresponds to the columns and width to the rows.

klemen1999 (approved):

PR #306, opened 2025-04-23
Export into multiple zip files

JSabadin:
## Purpose

- **Feature**: Added an option to split the exported dataset into multiple ZIP files.
- **Bugfix**: Fixed an issue where parsing a dataset with an empty split would result in a corrupted dataset.
- **Bugfix**: Fixed an issue where exporting a dataset containing files without annotations would cause those images to be ignored and excluded from the exported dataset.
- **Tests**: Added tests to cover the above bug fixes and the new feature for creating multiple ZIP files.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

**Note:** The dataset can be split into smaller parts, each compressed into a separate zip file. The original dataset can then be reconstructed by individually parsing each zip file.

### DIFF
--- a/luxonis_ml/data/__main__.py
+++ b/luxonis_ml/data/__main__.py
@@ -392,13 +392,39 @@ def export(
             help="Delete an existing `save_dir` before exporting.",
         ),
     ] = False,
+    task_name: Annotated[
+        Optional[str],
+        typer.Option(
+            "--task-name",
+            "-tn",
+            help=(
+                "Name of the single task to export. "
+                "Required when the dataset contains multiple tasks; "
+                "ignored if the dataset has exactly one task."
+            ),
+            show_default=False,
+        ),
+    ] = None,
+    max_zip_size_gb: Annotated[
+        Optional[float],
+        typer.Option(
+            ...,
+            "--max-zip-size-gb",
+            "-m",
+            help="Maximum size *per* ZIP archive, in gigabytes. "
+            "If the export grows beyond this limit, it is automatically split into "
+            "multiple parts (…_part1.zip, …_part2.zip, etc.). "
+            "Omit this option to create a single ZIP file, regardless of size.",
+            show_default=False,
+        ),
+    ] = None,
     bucket_storage: BucketStorage = bucket_option,
 ):
     save_dir = save_dir or dataset_name
     if delete_existing and Path(save_dir).exists():
         shutil.rmtree(save_dir)
     dataset = LuxonisDataset(dataset_name, bucket_storage=bucket_storage)
-    dataset.export(save_dir, dataset_type)
+    dataset.export(save_dir, dataset_type, max_zip_size_gb, task_name)
 
 
 @app.command()
--- a/luxonis_ml/data/__main__.py
+++ b/luxonis_ml/data/__main__.py
@@ -392,7 +392,7 @@ def export(
             help="Delete an existing `save_dir` before exporting.",
         ),
     ] = False,
-    task_name: Annotated[
+    task_name_to_keep: Annotated[
         Optional[str],
         typer.Option(
             "--task-name",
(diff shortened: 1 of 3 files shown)


tests/test_data/test_export.py
```diff
@@ -69,3 +72,55 @@ def test_dir_parser(
...
+    dataset.make_splits(ratios=(1, 0, 0), replace_old_splits=True)
+
+    loader = LuxonisLoader(dataset, view="train")
+    original_data = [img for img, _ in loader]
+
+    with subtests.test("Export with max_zip_size_gb=0.003"):
+        dataset.export(output_path=tempdir / "exported", max_zip_size_gb=0.003)
+
+        shutil.rmtree(tempdir / "exported" / dataset_name)
+        zip_files = sorted((tempdir / "exported").glob("*.zip"))
```
klemen1999, line 102: Let's also do a test where we call `LuxonisParser` on each zip seperately into the same `dataset_name`. So on the first call with first zip it should create a new dataset and then for the next zips it shold just add this data to an existing dataset. And the final dataset should be equivalent to the original one, before the export and parser

JSabadin: Previously, this wasn't possible since we only partitioned the dataset based on files. Now I've updated the logic to optionally split the dataset into smaller partitions and optionally compress them into zip files.

JSabadin: This makes it possible to parse the first zip file, then the second, and so on, reconstructing the original dataset from the smaller partitions.

luxonis_ml/data/__main__.py
```diff
@@ -392,13 +392,58 @@ def export(
...
+            "-m",
+            help=(
+                "Maximum size of each partition in GB. If the dataset"
+                " exceeds this size, it will be split into multiple partitions named {dataset_name}_part{partition_number}."
+                " Default is None, meaning the dataset will be exported as a single partition named {dataset_name}."
+            ),
+            show_default=False,
+        ),
+    ] = None,
+    zip_output: Annotated[
```
klemen1999, line 422: Let's have it as `--no-zip` store true type of flag so that by default we zip but we skip this step if the flag is present

tests/test_data/test_export.py
```diff
@@ -69,3 +71,58 @@ def test_dir_parser(
...
+            max_partition_size_gb=0.003,
+            zip_output=True,
+        )
+        zip_files = sorted((tempdir / "exported").glob("*.zip"))
+        assert len(zip_files) == 2
+
+    with subtests.test(
+        "Parse zip dirs into single dataset, having some splits empty"
+    ):
+        dataset = LuxonisParser(
```
klemen1999, line 109: we can do a foor loop over len(zip_files) for this and call LuxonisParser inside

tests/test_data/test_export.py
```diff
@@ -69,3 +71,58 @@ def test_dir_parser(
...
+            delete_local=False,
+            save_dir=tempdir,
+        ).parse()
+
+        dataset.make_splits(ratios=(1, 0, 0), replace_old_splits=True)
+
+        loader = LuxonisLoader(dataset, view="train")
+
+        new_data = [img for img, _ in loader]
+        assert len(new_data) == len(original_data)
```
klemen1999, line 128: Let's also add previous test (unzip into same dir and then call LuxonisParser on it) back in, it is a valid usecase that we also want to support

tests/test_data/test_export.py
```diff
@@ -69,3 +71,58 @@ def test_dir_parser(
...
+):
+    url = f"{storage_url}/{url}"
+    dataset = LuxonisParser(
+        url,
+        dataset_name=dataset_name,
+        delete_local=True,
+        save_dir=tempdir,
+    ).parse()
+
+    dataset.make_splits(ratios=(1, 0, 0), replace_old_splits=True)
```
klemen1999, line 92: We should also make an almost identical test function but without grouping all data into one split and instead making regular splits (as in leaving them like they were parsed), then zipping into multiple zips and then parsing them out into a new dataset that should have length of each split the same at the end

PR #318, opened 2025-05-08
Enable live reload of `environ` via proxy

JSabadin:
### PR Description
- Wrap `Environ` in a proxy so each attribute access clears cache and reloads from `.env`/`os.environ`  

**Example**  
_Before:_  
```python
from luxonis_ml.utils.environ import environ
import os
os.environ["POSTGRES_HOST"] = "new-db"
print(environ.POSTGRES_HOST)  # still returns old value
```
_after:_  

```python
from luxonis_ml.utils.environ import environ
import os
os.environ["POSTGRES_HOST"] = "new-db"
print(environ.POSTGRES_HOST)  # now prints "new-db"
```

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_ml/utils/environ.py
+++ b/luxonis_ml/utils/environ.py
@@ -1,6 +1,6 @@
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 from pydantic import model_serializer
 from pydantic_settings import BaseSettings, SettingsConfigDict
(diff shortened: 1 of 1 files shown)


klemen1999 (approved):
LGTM, CC: @kozlov721 could you check as well

kozlov721 (approved):

PR #330, opened 2025-06-05
Feat: Multi-Input Support

JSabadin:
## Purpose

This PR introduces support for **multi-input datasets**. Previously, dataset creation was constrained to a single image source per entry. Now, datasets can include multiple input files for each data point.

For example, a data record can be yielded like this:

```python
{
    "files": {
        "img_left": "path/to/img_left.png",
        "img_right": "path/to/img_right.png",
        "img_middle": "path/to/img_middle.png"
    },
    "task_name": "detection",
    "annotation": {
        "class": "person",
        "boundingbox": {
            "x": 0.1,
            "y": 0.1,
            "w": 0.3,
            "h": 0.4
        }
    }
}
```
In the loader, you specify the desired color space per input source using the `color_space` argument, for example:

```python
color_space = {
    "img_left": "GRAY",
    "img_right": "GRAY",
    "img_middle": "RGB"
}

  loader = LuxonisLoader(
        dataset,
        height=512,
        width=512,
        augmentation_config=augs,
        color_space=color_space,
    )
```

The keys of the image dictionary returned by the loader correspond to the source names defined in the `files` dictionary when yielding data records during dataset creation.

If a single image is being used per record, the behaviour is the same as before. 

When iterating over the loader:

```python
for img_dict, labels in loader:
    # img_dict will look like:
    # {
    #     "img_left": np.ndarray of shape (512, 512, 1),
    #     "img_right": np.ndarray of shape (512, 512, 1),
    #     "img_middle": np.ndarray of shape (512, 512, 3)
    # }
```

## Specification

Linking multiple sources into a single logical sample is handled via `group_id`s. These are computed by hashing the concatenation of sorted UUIDs in each group, ensuring deterministic grouping of related images.

The updated loader is backward-compatible with datasets built using previous versions. Existing datasets will load correctly with the new format, and datasets created under this version will be compatible with older versions of `luxonis-ml`.

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


Haricharan0310:
Hi @Username, I noticed you’ve been keeping docs up-to-date, always good to see.

I built a small LLM-based tool for maintaining high-level docs (SDK guides, API references, READMEs, tutorials). It auto-updates them on PRs to prevent drift.

No pressure at all, but thought it might be useful: https://github.com/marketplace/deepdocsai

klemen1999 (approved):
LGTM, thanks

PR #346, opened 2025-07-14
Remove Invisible Keypoints

JSabadin:
## Purpose

**Problem:**  
During a sequence of spatial transforms, keypoints could move outside the image bounds (for example, after a shift) and were not removed. A subsequent transform (for example, a rotation) might then bring those off-image points back inside the frame—only to have them point into empty (padded) areas.

**Solution:**  
We fixed this by setting  
```python
A.KeypointParams(format='xy', remove_invisible=True)
```

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_ml/data/augmentations/albumentations_engine.py
+++ b/luxonis_ml/data/augmentations/albumentations_engine.py
@@ -405,9 +405,7 @@ class AlbumentationsEngine(AugmentationEngine, register_name="albumentations"):
                 "bbox_params": A.BboxParams(
                     format="albumentations", min_visibility=min_bbox_visibility
                 ),
-                "keypoint_params": A.KeypointParams(
-                    format="xy", remove_invisible=False
-                ),
+                "keypoint_params": A.KeypointParams(format="xy"),
                 "additional_targets": self.targets
                 if is_custom
                 else targets_without_instance_mask,


luxonis_ml/data/__main__.py
```diff
@@ -349,7 +345,12 @@ def inspect(
     if aug_config is not None:
         h, w, _ = loader[0][0].shape
         loader.augmentations = loader._init_augmentations(
-            "albumentations", aug_config, h, w, not ignore_aspect_ratio
+            "albumentations",
```
klemen1999, line 348: Nitpick: Let's use explicit param assignment just in case order is switched or something new is added down the line. So h=h,w=w,...

luxonis_ml/data/augmentations/albumentations_engine.py
```diff
@@ -625,6 +640,26 @@ def postprocess(
...
+        shape = kwargs.get("shape")
+        if shape is None:
+            raise ValueError(
+                "Shape must be provided in kwargs to mark invisible keypoints."
+            )
+        h, w = shape[:2]
+        kps = keypoints.copy()
+        xs, ys = kps[:, 0], kps[:, 1]
+        oob = (xs < 0) | (ys < 0) | (xs >= w) | (ys >= h)
+        kps[oob, -1] = 0.0
```
klemen1999, line 660: Nitpick: lets use 2 instead of -1 here, makes more sense when reading the code

klemen1999 (commented):
Generally LGTM
Can we benchmark this a bit if it for some reason adds significant processing time?

PR #351, opened 2025-07-20
Add class reordering in dataset

sokovninn:
## Purpose
Adds class reordering functionality to the LuxonisDataset to allow users to specify custom class ordering for tasks. This enables control over class-to-index mapping which is important for model consistency and deployment scenarios where specific class ordering is required.

## Specification
- Implemented `set_class_order_per_task()` method with validation:
  - Checks if task names exist in dataset
  - Validates that provided classes match dataset classes exactly
  - Only reorders when class order differs from current order
  - Provides clear warning messages when reordering occurs
- Class reordering doesn't update dataset metadata on disk

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable


klemen1999 (commented):
LGTM but let's add some tests for it please

klemen1999 (approved):

PR #374, opened 2025-11-10
Extend dataset export functionality to all different DatasetTypes

dtronmans:
## Purpose
Go from only being able to export with DatasetType.NATIVE to all DatasetType entries.

## Specification
- Every current implemented parser now has a matching exporter.
- The logic for preparing the LDF (shared across all exporters), transforming it and saving are all separated
- BaseExporter defined to be implemented by different exporters, in a very similar way that BaseParsers defines an interface for different parsers

## Testing & Validation

The first way I tested the correctness of the functionality is to parse datasets, export them using the exporters then open them with different tools like Roboflow and FiftyOne:

parsed_fiftyone
parsed_dataset

The second and more precise way is by writing tests in tests/test_data/test_parse_export_equivalence.py, in which the following is tested:

1) Parse dataset collected from Roboflow, Google Cloud, etc... and save as prev_dataset
2) Export the dataset using defined exporters
3) Parse this exported dataset again and save as new_dataset
4) Assert that previous and new dataset are exactly the same. For bounding boxes, even though I try to save with as much precision as possible (.12f), we sometimes lose precision with repeated parsing and exporting so I assert that the normalized boxes are almost equal by rounding to 2 decimal places.

### DIFF
--- a/luxonis_ml/data/parsers/luxonis_parser.py
+++ b/luxonis_ml/data/parsers/luxonis_parser.py
@@ -50,6 +50,8 @@ class LuxonisParser(Generic[T]):
         DatasetType.SOLO: SOLOParser,
         DatasetType.NATIVE: NativeParser,
         DatasetType.YOLOV8BOUNDINGBOX: YOLOv8Parser,
+        DatasetType.YOLOV8INSTANCESEGMENTATION: YOLOv8Parser,
+        DatasetType.YOLOV8KEYPOINTS: YOLOv8Parser,
     }
 
     def __init__(


luxonis_ml/data/exporters/exporter_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        splits: dict[str, Any],
+        processed_df: pl.DataFrame,
+        grouped_image_sources: pl.DataFrame,
+    ):
+        self.splits = splits
+        self.processed_df = processed_df
+        self.grouped_image_sources = grouped_image_sources
+
+    @classmethod
+    def from_dataset(cls, ldf: "LuxonisDataset") -> "PreparedLDF":
```
kozlov721: It would be better to rename `ldf` to `dataset` to be consistent with the rest of the codebase.

luxonis_ml/data/exporters/exporter_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        )
+
+
+@dataclass(frozen=True)
+class ExporterSpec:
+    cls: type["BaseExporter"]
+    kwargs: dict
+
+
+class ExporterUtils:
```
kozlov721: These static methods can be just normal functions.

luxonis_ml/data/parsers/luxonis_parser.py
```diff
@@ -49,7 +49,7 @@ class LuxonisParser(Generic[T]):
         DatasetType.SEGMASK: SegmentationMaskDirectoryParser,
         DatasetType.SOLO: SOLOParser,
         DatasetType.NATIVE: NativeParser,
-        DatasetType.YOLOV8: YOLOv8Parser,
+        DatasetType.YOLOV8BOUNDINGBOX: YOLOv8Parser,
```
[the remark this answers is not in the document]
dtronmans: Our YOLOV8 parser incorrectly parses keypoints, segmentation masks and bounding boxes annotation-by-annotation, while in practice Ultralytics only accepts one annotation type (either bounding boxes, or instance segmentation or keypoints). Therefore our parser is more lenient than what Ultralytics expects, and if we allow users to export to all YOLOV8 formats simultaneously then this exported dataset will not be parsable by Ultralytics.

PR #379, opened 2026-01-06
Feat/fiftyone classification

dtronmans:
## Purpose
Expand the parse and export suite with more classification formats that are accepted by FiftyOne. These include three new formats:

- Flat collection of folders, each folder corresponding to a class name: before, `classification_directory_parser` was able to do this if these sub-folders were divided into train/valid/test/ subdirectories, now this works with a flat structure. The splits are generated at parse-time randomly.
- train/validation/test subdirectories with a `data` folder and a `labels.json` residing in each
- Flat structure: without the splitting of train/validation/test, just a flat structure with a `data` folder where the images reside and a `labels.json` file. Again the splits are randomly assigned

I did not have to implement the mechanism that randomly assigns splits for flat directory structure because this is handled by the LuxonisDataset `make_splits()` method which only assigns splits to unassigned ids. Here the default ratio is `{"train": 0.8, "val": 0.1, "test": 0.1}`

- Implemented the corresponding ClassificationDirectoryExporter

## Testing & Validation

After parsing then exporting, then opening a session with the exported datasets:

MNIST:
mnist

Caltech101:
caltech101

Imagenet-sample:
imagenetsample

### DIFF
--- a/luxonis_ml/data/parsers/classification_directory_parser.py
+++ b/luxonis_ml/data/parsers/classification_directory_parser.py
@@ -53,7 +53,7 @@ class ClassificationDirectoryParser(BaseParser):
                 "valid",
                 "test",
                 "val",
-                "valdation",
+                "validation",
                 "images",
                 "labels",
             }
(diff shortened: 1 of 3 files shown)


klemen1999 (commented):
Structurally LGTM CC: @ptoupas if you could test out your usecase

klemen1999 (approved):

PR #382, opened 2026-01-20
Feat/make splits cli

dtronmans:
## Purpose
Right now it's not exactly obvious how re-shuffling images, selecting subsets etc can be done and it would be good to give more flexibility regarding dataset manipulation at parse-time. In this PR the `--split-ratio` argument is added to the CLI parse command, such that at parse-time/right after parsing either splits or subsets can be defined. `make_splits()` remains exactly the same, so if someone is used to using the Python API `make_splits()` instead of using this flag, the original functionality remains. The original functionality also remains if someone does not specify --split-ratio.

## Testing & Validation
When calling `luxonis_ml data parse --help`, the following flag is added:

gh_helpmessage

Parsing with ratios `0.5,0.1,0.4`:

gh_050104

Parsing with counts (subsets of the original splits) `5,60,0`:

gh_5600

### DIFF
--- a/luxonis_ml/data/parsers/base_parser.py
+++ b/luxonis_ml/data/parsers/base_parser.py
@@ -280,15 +280,23 @@ class BaseParser(ABC):
         @rtype: LuxonisDataset
         @return: C{LDF} with all the images and annotations parsed.
         """
-        train_dir = dataset_dir / "train"
-        if not train_dir.exists() or not train_dir.is_dir():
-            existing_dirs = [
-                d.name for d in dataset_dir.iterdir() if d.is_dir()
-            ]
-            raise ValueError(
-                f"Train split not found in dataset. "
-                f"Expected a 'train' directory but found: {existing_dirs}."
-            )
+        # Skip train directory check for parsers that use images/labels
+        # subdirectory structure (YoloV6, YOLOv8) instead of train/valid/test
+        # at root level
+        skip_train_check = self.__class__.__name__ in (
+            "YoloV6Parser",
+            "YOLOv8Parser",
+        )
+        if not skip_train_check:
+            train_dir = dataset_dir / "train"
+            if not train_dir.exists() or not train_dir.is_dir():
+                existing_dirs = [
+                    d.name for d in dataset_dir.iterdir() if d.is_dir()
+                ]
+                raise ValueError(
+                    f"Train split not found in dataset. "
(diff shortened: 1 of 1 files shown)


luxonis_ml/data/__main__.py
```diff
@@ -55,6 +56,54 @@ def complete_dataset_name(incomplete: str):
 )
 
 
+def parse_split_ratio(
```
klemen1999: Let's put this + all other functions that are not @app.command to a separate util function (e.g. `utils/general_utils.py`) to keep the `__main__` clean

dtronmans: Regarding a final re-test, Petros gave me a few comments today offline which I'm fixing right now:

1) If there are more requested images than images available in a flat dataset structure (for example, imagenet-sample is flat with 1000 images and we request [2000, 3000, 2000]) then instead of defaulting to [0.8, 0.1, 0.1], fill the splits by prioritizing the splits with the highest number of images requested.
2) Disable [split_val_to_test ](https://github.com/luxonis/luxonis-ml/blob/main/luxonis_ml/data/parsers/coco_parser.py#L129) if counts are specified. Personally I'm not sure why split_val_to_test is implemented because it seems to alter the distribution behavior of images in splits only for the COCO parser.
3) The COCO parser should count the number of images it encounters without annotations [here](https://github.com/luxonis/luxonis-ml/blob/main/luxonis_ml/data/parsers/coco_parser.py#L264) and warn the user in the end, something like "x images skipped because they did not have matching annotation entries". This is not really in the scope of the PR but I was thinking of adding it here

ptoupas:
I further tested the new flag, and it looks like the first point from [this comment](https://github.com/luxonis/luxonis-ml/pull/382#discussion_r2736019333) is now nicely addressed. The splits are created as expected, following the priorities we discussed offline with Dani.

However, running the following parse command:
```bash
luxonis_ml data parse /home/ptoupas/fiftyone/coco-2017 --name coco-2017-test --split-ratio '[200,200,10]'
```
now results in a `KeyError` in `luxonis-ml/luxonis_ml/data/parsers/coco_parser.py:235` when accessing the `annotations` key. It seems that when samples are requested from the COCO test split (where annotations are missing), they are not being properly skipped. Instead, the parser still attempts to read annotations and create the corresponding split in the LDF.

ptoupas (approved):
After a few iterations of offline discussions, and after testing again it LGTM.

PR #391, opened 2026-02-18
Changes to parsing behavior, documentation and tests

dtronmans:
## Purpose
- Allow keypoint annotations parsing for COCO using the `use_keypoint_ann` when test annotation json is 
- Revert previous change: previously I allowed all images to contain hard negatives (no ground truth labels) for dense estimation tasks (instance segmentation, detection and keypoints). Now just for coco-2017 keypoints we don't parse these images with no annotations, because over 50% of the images in the dataset don't have keypoints
- Documentation regarding parsing coco-2017 and imagenet-sample to make everything clearer
- Test structure and first few tests for parsing using CLI, we can add more over time depending on edge cases that appear or behavior we want to enforce.
- Added coco keypoints to `test_parse_export_equivalence` and added equivalency testing for semantic segmentation masks (realized I had it for instance segmentation but not semantic segmentation)

### DIFF
--- a/tests/test_data/utils.py
+++ b/tests/test_data/utils.py
@@ -136,9 +136,9 @@ class LDFEquivalence:
 
         if collector.__name__ == "collect_bbox_multiset":
             LDFEquivalence.multiset_equal_with_tolerance(prev, new, tol=0.02)
-        elif (
-            collector.__name__
-            == "collect_instance_segmentation_mask_overlap_multiset"
+        elif collector.__name__ in (
+            "collect_instance_segmentation_mask_overlap_multiset",
+            "collect_segmentation_mask_overlap_multiset",
         ):
             # Combine all masks per image and assert there is spatial overlap.
             def _combine_masks(masks: list[np.ndarray]) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


luxonis_ml/data/parsers/README.md
```diff
@@ -62,3 +75,246 @@ This mode:
 - Take 500 random samples from the original train split
 - Take 100 random samples from the original val split
 - Take 0 samples from the test split
+
+## Common datasets for evaluation
```
klemen1999: Nitpick: Let's put this whole section one level up so its not nested under LuxonisParser but instead in its own section.

luxonis_ml/data/parsers/README.md
```diff
@@ -62,3 +75,246 @@ This mode:
...
+
+# With a subset of samples
+luxonis_ml data parse ./coco-2017 --split-ratio 1000,200,100
+```
+
+### Corrupted Image Handling
+
+The COCO-2017 train set contains a small number of images known to cause issues. The parser automatically filters these out when parsing the train split. No user action is required.
+
+## Imagenet-sample Dataset Parsing
```
ptoupas: Lets don't make this specific to the [ImageNet Sample](https://docs.voxel51.com/dataset_zoo/datasets/imagenet_sample.html#imagenet-sample) dataset.
Lets just talk about the [ImageNet 2012](https://docs.voxel51.com/dataset_zoo/datasets/imagenet_2012.html#imagenet-2012) (or just ImageNet) dataset instead since `ImageNet Sample` is just a subset of that.

ptoupas (approved):
Left a couple of minor comments, but apart from that it LGTM.

PR #393, opened 2026-02-23
Parse zip with top level directory

dtronmans:
## Purpose
Make it possible to parse a zip file with a top-level directory instead of immediate sub-directories.

## Specification
- Assumes there is only one directory at the root of the zip and nothing else. I filter away __MACOSX, Thumbs.db and desktop.ini as well as anything that could start with `.` (.DS_STORE, .cache) in case these are created with the top-level dir when the zip file is created.
- The only possible unintended change in parsing is that `ClassificationDirectoryParser `could falsely win on wrapped archives because it accepts generic class-folder layouts. To avoid that, we unwrap a single top-level folder only when it has explicit dataset-root markers (eg train/valid/test, images/labels, annotations.json, labels.json, data.yaml) or when that inner folder is recognized by any non-CLSDIR parser.
- `test_zip_layout_equivalence.py`: for different zip datasets, parse the original (flat) zip, parse the zip with an added top-level directory, and parse the extracted zip folder and assert that all 3 return the same LDFs.
- For `test_zip_layout_equivalence.py`, I used the helper methods from test_parse_export_equivalence.py to assert that the resulting annotations are the same, so since they are used in two tests I moved them to `utils.py`

## Testing & Validation

imagenet-sample zip file with top-level dir, parser detects the correct format:
correctimagenet

coco-2017 zip file with top-level dir, parser detects the correct format:
correctcoco

### DIFF
--- a/luxonis_ml/data/parsers/luxonis_parser.py
+++ b/luxonis_ml/data/parsers/luxonis_parser.py
@@ -172,10 +172,10 @@ class LuxonisParser(Generic[T]):
         if not only_entry.is_dir():
             return unzip_dir
 
-        # only unwrap when the inner directory clearly
-        # looks like a dataset root, not an arbitrary folder.
-        # this is because ClassificationDirectoryParser accepts
-        # any directory whose children are subdirectories (class names)
+        # Only unwrap when the inner directory clearly looks like a
+        # dataset root, not an arbitrary folder.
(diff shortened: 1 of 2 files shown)


klemen1999 (approved):
Tested on a few examples and LGTM, ty

PR #403, opened 2026-03-31
LDF equivalence method

dtronmans:
## Purpose
- Compare two LDFs to assert that they have equivalent images and annotations
- Image order or file renaming are not taken into account because images are matched using their hashes, so LDFs with the same image set and equivalent annotations but in a different order will be considered equivalent
- Uses previous utils defined for `test_parse_export_equivalence`

## Specification
Usage: take as input either a LuxonisDataset object or a string. Both of these blocks are accepted:

```
ldf_a = LuxonisDataset("parkinglot-a")
ldf_b = LuxonisDataset("parkinglot-b")

print(ldf_equivalent(ldf_a, ldf_b))
```

```
ldf_a = "parkinglot-a"
ldf_b = "parkinglot-b"

print(ldf_equivalent(ldf_a, ldf_b))
```

Using __eq__ between two LDFS:

```
ldf_a = "parkinglot-b"
ldf_b = "parkinglot-a"

print(LuxonisDataset(ldf_a) == LuxonisDataset(ldf_b))
```

Between an LDF and a string:

```
ldf_a = "parkinglot-b"
ldf_b = "parkinglot-a"

print(LuxonisDataset(ldf_a) == ldf_b)
```

## Testing & Validation
- Refactored the utils defined in `test_parse_export_equivalence` by moving them to `luxonis-ml[data]` utils
Three assertions in `test_ldf_equivalence`:
  - Same datasets return True for equivalence
  - Different datasets return False
  - Slightly modifying the keypoint annotations of a dataset and comparing it with its previous version returns False


kozlov721 (approved):
LGTM, but could be good to also define `LuxonisDataset.__eq__`

klemen1999 (approved):
Generally LGTM, didn't review the LDFEquivalence class implementation that closesly but I assume that is fine since it was already "tested" in tests before.

PR #405, opened 2026-04-03
Classification and metadata visualizer

dtronmans:
## Purpose
Classification and metadata/ tasks did not show any labels when inspected using `luxonis_ml data inspect <dataset-name>`, this PR adds visualization for both.

This was also tested with `luxonis_train inspect --config <config-yaml>`

If `luxonis_train inspect --config <...>` is called on a dataset with multiple metadata labels, but the `task_name` is set in the predefined model config, then only metadata labels pertaining to this task name are shown.

Example with `luxonis_train inspect ...` and predefined model head has `task_name` cars:

cars_only

`task_name` vehicles:

only_vehicles

Note: the numbers are shown because `luxonis-train`'s loader encodes string metadata as character-coded arrays before passing it to the `inspect`, this will be fixed in a separate `luxonis-train` PR

Examples with `luxonis_ml data inspect`:

classification_only

metadataonly

Both metadata labels and classification (synthetic dataset for the sake of the example);

### DIFF
--- a/luxonis_ml/data/utils/visualizations.py
+++ b/luxonis_ml/data/utils/visualizations.py
@@ -591,10 +591,12 @@ def visualize(
         if isinstance(value, np.generic):
             value = value.item()
 
-        if task in metadata_mappings:
+        if task in metadata_mappings and isinstance(
+            value, (int, str, bytes, bytearray)
+        ):
             try:
                 return metadata_mappings[task].inverse[int(value)]
-            except (KeyError, TypeError, ValueError):
+            except (KeyError, ValueError):
                 pass
 
         if isinstance(value, float):
--- a/luxonis_ml/data/utils/visualizations.py
+++ b/luxonis_ml/data/utils/visualizations.py
@@ -741,11 +741,10 @@ def visualize(
     text_lines: list[str] = []
     if classification_only and len(classification_labels) == 1:
         _, class_names = classification_labels[0]
-        prefix = "Class" if len(class_names) == 1 else "Classes"
-        text_lines.append(f"{prefix}: {', '.join(class_names)}")
+        text_lines.extend(["Classes:", ", ".join(class_names)])
     elif classification_only and classification_labels:
         text_lines.extend(
-            ["Classification labels"]
+            ["Classification labels:"]
             + [
                 f"{task_name}: {', '.join(class_names)}"
                 for task_name, class_names in classification_labels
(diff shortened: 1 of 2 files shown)


luxonis_ml/data/utils/visualizations.py
```diff
@@ -566,4 +730,39 @@ def create_mask(
 
         images[image_name] = curr_image
 
-    return concat_images(images)
+    output = concat_images(images)
+
+    classification_only = all(
```
klemen1999: I'd remove this gate and just always display the classification and metadata (when they are not empty)

luxonis_ml/data/utils/visualizations.py
klemen1999, line 1: Nitpitck on the function definition order: Let's have the order of the functions in the order that they are used + their importance. So in this case it would be:
- `concat_images` (most important function, used elsewhere directly)
- `append_text_blocks` - used by `concat_images`
- `wrap_text` - used by `append_text_blocks`

This makes the code easier to read and maintain IMO

klemen1999 (approved):
Overall LGTM

luxonis_ml/data/utils/visualizations.py
dtronmans: This becomes less of a pure "classification visualizer" then because there are still classification labels written when there are instance segmentation or detection annotations, for example. So in this case there might be confusion regarding the bottom block being either an indicator for classification label for the entire image or just which classes are present as objects in the image. For example this is what it looks like when I remove the gate:

classes_non_classification
example

klemen1999: But in both of these case the dataset can be used also for training classification model right? The classification annotations are actually there. I think in the `data inspect` "mode" I'd like to see every annotation that is present on the image which in these cases includes also the class - I agree that it is not super useful but it is there and present.

dtronmans: True, it can still be trained for classification and in this case the labels are just the present objects. Good point I will remove the gate.

luxonis_ml/data/utils/visualizations.py
dtronmans: Addressed in 5ff5c3a

luxonis_ml/data/utils/visualizations.py
dtronmans: Addressed in 5ff5c3a

klemen1999 (approved):

PR #406, opened 2026-04-08
use_for_resizing augmentation is wrapped in a OneOf with default resize if p < 1

dtronmans:
## Purpose
- Allow a passed-down `use_for_resizing` with p < 1 for an augmentation, but then wrap it inside a p = 1 `OneOf` transform where the other transform is a default-behavior resize transform with a `p - 1` probability
- Even though the probability passed from the luxonis-train is already a float, the validation step `isinstance(resize_probability, bool) or not isinstance(resize_probability, (int, float)):` had to be added for PyRight
- [Corresponding luxonis-train PR](https://github.com/luxonis/luxonis-train/pull/368) that removes the p == 1 restriction/override
- No behavior change from previous code for `p == 1` `use_for_resizing` augmentation, this OneOf is only created if `p < 1.0`

### DIFF
--- a/tests/test_data/test_augmentations/test_special.py
+++ b/tests/test_data/test_augmentations/test_special.py
@@ -1,4 +1,6 @@
+import albumentations as A
 import numpy as np
+import pytest
 
 from luxonis_ml.data import AlbumentationsEngine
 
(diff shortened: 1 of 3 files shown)


tests/test_data/test_augmentations/test_special.py
klemen1999, line 1: Let's just add a test where keep_aspect_ratio=True and custom resize is used + p<1. Just to see that LetterboxResize is correctly used in this case.

klemen1999 (approved):
LGTM

tests/test_data/test_augmentations/test_special.py
dtronmans: Addressed in 692eba9

klemen1999 (approved):

PR #410, opened 2026-04-29
Specify the stage (train/val/test) instead of generic `is_validation_pipeline`

dtronmans:
## Purpose
Unfortunately it was not previously possible for the `AlbumentationsEngine` to distinguish between train, val and test, because its only demarcation was train vs. not-train through the `is_validation_pipeline`. In this PR `is_validation_pipeline` is replaced with `pipeline_stage` to make it possible to treat train val and test separately

However it's still possible that somewhere in other repos/locally there is an `AlbumentationsEngine(..., is_validation_pipeline=True)` defined so in the loader we still pass-through this argument without crashing just for backwards-compatibility sake.

Also allow `apply_on_stages` on augmentation config items (default ["train"]) so specific augmentations can be enabled for val and/or test as needed.

## Testing & Validation

Backwards-compatibility: manually tested that this does not crash/return an error:

```
if __name__ == "__main__":
    try:
        engine = AlbumentationsEngine(
            256,
            256,
            {"/classification": "classification"},
            {"/classification": 1},
            ["image"],
            [{"name": "Normalize"}],
            is_validation_pipeline=True,
        )
        print(type(engine).__name__)
    except TypeError as exc:
        print(f"TypeError: {exc}")
```


luxonis_ml/data/augmentations/base_engine.py
```diff
@@ -28,7 +30,8 @@ def __init__(
         source_names: list[str],
         config: Iterable[Params],
         keep_aspect_ratio: bool,
-        is_validation_pipeline: bool,
+        is_validation_pipeline: bool | None = None,
```
klemen1999, line 33: Can we at some stage (not sure if this is in base_engine or in luxonis_loader) detect if `is_validation_pipeline` has been explicitly set and if it has issue a deprecation warning?
Is [this](https://github.com/luxonis/luxonis-ml/blob/db14506677abc4b3db1a8e5c90e333814f7a8c79/luxonis_ml/data/datasets/luxonis_dataset.py#L1281) decorator maybe usefule here?

dtronmans: Yes nice idea I did not know about this decorator. I added the deprecation note now, here is what is logged when `AlbumentationsEngine(..., is_validation_pipeline=True)` is initialized:

```
DeprecationWarning: Argument
         'is_validation_pipeline' in function '__init__' is deprecated and will be removed in future versions. Use 'pipeline_stage'                 
         instead. Use `pipeline_stage='train'`, `'val'`, or `'test'` instead. 
```

klemen1999 (approved):

PR #415, opened 2026-05-21
More permissive COCO parser: allow strings for bbox values

dtronmans:
## Purpose
- Accept strings in bbox values (use the new warnings collector to emit a warning if the string is un-parsable to float like "113a")

## Testing & Validation
Parsed a dataset locally with `--split-ratio "[1.0, 0.0, 0.0]"`, exported to COCO Roboflow format, parsed in Roboflow UI, exported from there (still with the 1.0/0/0 split) and parsed using the COCO parser.

## AI Usage

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved COCO annotation validation: bounding box values are now verified to be numeric and finite; invalid annotations are automatically skipped and reported as parsing issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luxonis/luxonis-ml/pull/415?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


luxonis_ml/data/utils/enums.py
```diff
@@ -51,6 +51,7 @@ class COCOFormat(str, Enum):
 class ParserIssue(Enum):
     MISSING_IMAGE = "missing_image"
     COCO_ISCROWD = "coco_iscrowd"
+    COCO_NON_NUMERIC_BBOX = "coco_non_numeric_bbox"
```
klemen1999: Let's name this in a more general way e.g. `NON_NUMERIC_ANNOTATION` as I can see this being a problem with other formats as well where some bad annotations could come up and should be skipped due to them being non numeric. We don't really check for this ATM but later on I can see that we'll have to essentially check every cast

dtronmans: True, addressed in [9006f16](https://github.com/luxonis/luxonis-ml/pull/415/commits/9006f16442f8897aaf54d05a5eab670be3c59ab0)

luxonis_ml/data/parsers/coco_parser.py
```diff
@@ -332,7 +336,17 @@ def generator() -> DatasetIterator:
                             "counts": coco_seg["counts"],
                         }
 
-                    x, y, w, h = ann["bbox"]
+                    try:
+                        x, y, w, h = (float(value) for value in ann["bbox"])
```
klemen1999, line 337: We should also check here if value is finite because currently technically if string value is `"inf"` the `float(value)` passes without issues but annoation can't actually be inifinte.

klemen1999 (commented):
Let's remove the "allow train/ split only" part out of this PR and address it in a separate one. I think that one requires a bit more thought and a wholistic solution to maybe even generally allow directory to have any subset of splits.

luxonis_ml/data/parsers/coco_parser.py
dtronmans: Thanks, addressed both in [9d13c94](https://github.com/luxonis/luxonis-ml/pull/415/commits/9d13c943c139b44bff64d89d49993cc8460d7336) will change the PR title and PR comment now

klemen1999 (approved):

PR #418, opened 2026-06-07
CLI: migration to `cyclopts`

kozlov721:
## Purpose

Makes the CLI in `luxonis-ml` consistent with `luxonis-train` and `modelconverter`.

## Specification

- Replaced `typer` with `cyclopts`
- Deprecated `luxonis_ml data parse --split-ratios`
  - `--train`, `--val`, and `--test` instead
  - `luxonis_ml data parse -sr "[0.8,0.1,0.1]"` $\rightarrow$ `luxonis_ml data parse --train 0.8`
- Added an option to print the `buildinfo.json` to `luxonis_ml archive inspect` (`-b`)
- `luxonis_ml data delete` can now delete multiple datasets at once
  - And supports `-y` flag to automatically confirm the deletions 

### Open questions
As opposed to `typer`, `cyclopts` does not yet support dymanic completions ([Issue #641](https://github.com/BrianPugh/cyclopts/issues/641)). This means that shell completions for dataset names will no longer work.
This gives us 3 options:
  1. Implement dynamic completion ourselves
  1. Wait with merge until supported upstream
  1. Merge anyways and add support when available

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **Refactor**
  * Migrated the dataset, NN archive, filesystem utilities, benchmark, and main CLIs to a unified command-line framework with consistent help/version and standardized subcommand behavior.
* **New Features / UX**
  * Dataset delete accepts multiple targets with a `--yes` bypass; parse supports `--train/--val/--test` splits (legacy ratio still supported).
  * Archive inspect adds `-b/--buildinfo`; filesystem `ls` supports `--type file|directory|all`.
* **Bug Fixes**
  * Improved validation for split inputs and missing resources; safer archive extraction and clearer output.
  * Updated push/pull/clone/merge/sanitize CLI options and constraints.
* **Tests / Chores**
  * Added CLI smoke tests for key commands; disabled docs job in CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,6 +48,7 @@ jobs:
   docs:
     needs: pre-commit
     runs-on: ubuntu-latest
+    if: false
 
     steps:
       - name: Checkout
--- a/luxonis_ml/data/__main__.py
+++ b/luxonis_ml/data/__main__.py
@@ -55,8 +55,8 @@ def info(
     """Print information about a dataset.
 
     Args:
-        name (str): Name of the dataset.
-        bucket_storage (BucketStorage): Storage type of the dataset.
+        name: Name of the dataset.
+        bucket_storage: Storage type of the dataset.
     """
     check_exists(name, bucket_storage)
     print_info(LuxonisDataset(name, bucket_storage=bucket_storage))
@@ -79,11 +79,11 @@ def delete(
     """Delete a dataset from local storage, remote storage, or both.
 
     Args:
-        names (str): Name(s) of the dataset to delete.
-        bucket_storage (BucketStorage): Storage type of the dataset.
-        local (bool): If True, delete the dataset from local storage.
-        remote (bool): If True, delete the dataset from remote storage.
-        yes (bool): If True, skip confirmation prompt and delete immediately.
+        names: Name(s) of the dataset to delete.
+        bucket_storage: Storage type of the dataset.
+        local: If True, delete the dataset from local storage.
+        remote: If True, delete the dataset from remote storage.
+        yes: If True, skip confirmation prompt and delete immediately.
     """
     if not local and not remote:
         print(
@@ -146,9 +146,9 @@ def ls(
     """List datasets.
 
     Args:
-        full (bool): If True, show full information about each dataset,
+        full: If True, show full information about each dataset,
             including classes and tasks.
-        bucket_storage (BucketStorage): Storage type of the dataset.
+        bucket_storage: Storage type of the dataset.
     """
     datasets = LuxonisDataset.list_datasets(bucket_storage=bucket_storage)
     table = Table(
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,6 +48,7 @@ jobs:
   docs:
     needs: pre-commit
     runs-on: ubuntu-latest
+    if: false
```
klemen1999, line 51: Why is this?

luxonis_ml/data/__main__.py
```diff
@@ -41,101 +40,116 @@
...
-        False,
-        "--local/--no-local",
-        help="Delete the dataset from local storage.",
-    ),
-    remote: bool = typer.Option(
-        False,
-        "--remote/--no-remote",
-        help="Delete the dataset from remote storage.",
-    ),
+    *names: str,
```
klemen1999, line 67: We should error out if no names are passed otherwise command like `luxonis_ml data delete --local` doesn't do anything but also doesn't error out

luxonis_ml/data/__main__.py
```diff
@@ -840,50 +804,37 @@ def pull(
...
-        str,
-        typer.Argument(
-            ..., help="Name of the new dataset.", show_default=False
-        ),
-    ],
-    push_to_cloud: Annotated[
+    name: str,
+    new_name: str,
+    *,
+    push: Annotated[
```
klemen1999, line 809: These types of breaking changes (e.g. push_to_cloud -> push, splits_to_cloe -> splits, ...) should be explicitly noted in the PR description. Check if there are any others as well.

luxonis_ml/data/__main__.py
```diff
@@ -840,50 +804,37 @@ def pull(
...
-            "-s",
-            help=(
-                "Comma separated list of split names to clone, "
-                "e.g. `-s val,test` or just `-s train`. "
-                "If omitted, clones all splits."
-            ),
-            show_default=False,
-        ),
+    bucket_storage: BucketStorageT = BucketStorage.LOCAL,
+    splits: Annotated[
```
klemen1999: What is the correct way to pass list[str] through CLI? Becuase if I do `-s "val,test"` this would be evaluated as `["val,test"]` right?
For this kinds of things I like having the help/docstrings a bit more explicit with maybe an example (like we had with the previous help text)

luxonis_ml/data/__main__.py
kozlov721: The CLI argument still has the same name. The function argument was called `push_to_cloud` but in the CLI it was aliased to `--push`, same for `splits_to_clone` etc.

luxonis_ml/data/__main__.py
kozlov721: That would be `... -s val -s test`. I'll add it to examples in the README

.github/workflows/ci.yaml
kozlov721: Just to temporarily disable the doc build tests until #424 is merged.

luxonis_ml/data/__main__.py
klemen1999: README is good but I think it would be also good if added directly to the docstring because in that case you have this example at hand when doing --help

klemen1999 (approved):

PR #424, opened 2026-06-16
Enhance documentation

kozlov721:
## Purpose

Improves the documentation and migrates the used style to google docstyle.

The rendered documentation can be accessed [here](https://luxonis.github.io/luxonis-ml/latest/).

## Specification

None / not applicable

## Dependencies & Potential Impact

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Submitted code was reviewed by a human: YES/NO

The author is taking the responsibility for the contribution: YES/NO

## Summary by CodeRabbit

* **New Features**
  * Added GitHub Pages deployment for API docs (latest + versioned) and local doc build instructions.
  * Exposed additional public APIs (e.g., `BatchTransform`, `NativeParser`) and expanded NN Archive public entry points.
  * Updated `push_to_cloud` to allow omitting bucket storage (uses the dataset default).

* **CI / Chores**
  * Enabled docs generation in CI and updated the docs pipeline; tests now wait for docs.
  * Removed the dead-fixtures check; adjusted semgrep conditions and pre-commit hooks.

* **Documentation**
  * Replaced multiple module READMEs with deprecation redirects and refreshed large-scale docstrings/README links.

PR #426, opened 2026-06-18
Small breaking changes - docs pre-requirement

kozlov721:
## Purpose

Moves some breaking code changes out of #424.

## Specification

- Private method and attributes renamed with a leading underscore
- Fixed many typos and added a check for typos to `pre-commit`

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **New Features**
  * Added a CLI option to log all parser warnings during data import.
  * Enabled automatic dataset/plugin registration at startup.
  * Added typo-checking to pre-commit validation.

* **Bug Fixes**
  * Corrected symmetric keypoint flip/transpose augmentation names and references.
  * Improved dataset/parsing robustness by failing fast with clearer errors when images/masks can’t be loaded.
  * Tightened identifier validation for annotations to prevent invalid task/label names.

* **Documentation**
  * Fixed multiple typos in READMEs, examples, and docstrings.


luxonis_ml/utils/logging.py
```diff
@@ -95,7 +95,7 @@ def _custom_warning_handler(
         category: type[Warning],
         filename: str,
         lineno: int,
-        _file: str | None = None,
+        file: str | None = None,
```
klemen1999, line 98: Where is this param even used? I don't see it used in this function..

kozlov721: It's not, this is just part of the API for warnings handler.

klemen1999: Was it `_file` before to not be confused with "main" `file` variable on the top function?

kozlov721: No it was named `_file` only because some linters will take the hint and won't complain about an unused argument. But in general changing the names of arguments of monkey-patched functions is not great so I changed it back to `file`. It doesn't really matter, we just never use the leading underscore for unused arguments anywhere else.

klemen1999 (approved):

klemen1999 (approved):
Generally LGTM.

PR #434, opened 2026-06-26
Dataset Record Metadata

kozlov721:
## Purpose

Add record-level dataset metadata support while preserving compatibility with existing loader consumers and older annotation parquet files.

## Specification

- Store sample metadata in parquet rows.
- Expose loader metadata via LoaderOutput.metadata.
- Keep LoaderOutput tuple-compatible for existing two-value unpacking.
- Add default metadata handling for old parquet schemas.
- Add `--show-metadata` to opt into metadata printing during dataset inspection.
- By default `LuxonisLoader` adds `"filenames"` field to the sample metadata (can be disabled by setting `autopopulate_metadata=False`)

### Record metadata

A new field `"sample_metadata"` is added to the schema of `DatasetRecord`.

```python
yield {
    "file": "images/img_0.jpg",
    "sample_metadata": {
        "capture_id": "frame-0001",
        "tags": ["rain", "night"],
    },
    "annotation": {
        "class": "car",
        "boundingbox": {"x": 0.1, "y": 0.2, "w": 0.3, "h": 0.4},
    },
}
```

### Batch augmentation metadata

For batch augmentations such as `MixUp` or `Mosaic4`, we take the metadata from the first record and add a new field `"batch_augmentation_metadata"` containing metadata of the full batch.  

```python
{
    "filenames": {"image": "first.jpg"},
    "tags": ["night"],
    "batch_augmentation_metadata": [
        {
            "input_index": 0,
            "sample_metadata": {
                "filenames": {"image": "first.jpg"},
                "tags": ["night"],
            },
        },
        {
            "input_index": 1,
            "sample_metadata": {
                "filenames": {"image": "second.jpg"},
                "tags": ["day", "rain"],
            },
        },
    ],
}
```

## Testing & Validation

Added tests for:

- reordered metadata columns in loader dataframes
- mixed old/new parquet schemas
- appending to old-schema datasets
- push/pull metadata preservation
- native export/import metadata roundtrip

## AI Usage

Assisted-by: OpenAI:gpt-5 codex

## Summary

* **New Features**
  * Dataset records now include per-sample metadata across loading, augmentations, and native export/import.
  * The data inspection CLI can print per-sample metadata (enabled by default, toggleable).
* **Bug Fixes**
  * Parquet ingestion now consistently backfills missing or null metadata.
  * Dataset LDF compatibility now tolerates major-version differences.
* **Refactor**
  * Loader outputs are now a structured result that includes images, labels, and metadata.
* **Tests**
  * Added/updated coverage for metadata persistence and Parquet schema backfilling.


luxonis_ml/typing.py
```diff
@@ -41,11 +61,32 @@
...
 
-LoaderOutput: TypeAlias = LoaderSingleOutput | LoaderMultiOutput
-"""C{LoaderOutput} is a tuple containing either a single image as a
-C{np.ndarray} or a dictionary mapping image names to C{np.ndarray},
-along with a dictionary of task group names and their annotations as
-L{Annotations}."""
+
+@dataclass(slots=True)
+class LoaderOutput(
+    tuple[np.ndarray | dict[str, "np.ndarray"], Labels]
```
kozlov721, line 67: For the sake of backwards compatibility (so `for images, labels in loader` does not break).

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -473,8 +501,32 @@ def _load_with_augmentations(self, idx: int) -> LoaderMultiOutput:
...
-        return self.augmentations.apply(loaded_anns)
+        loaded_anns: list[tuple[dict[str, np.ndarray], Labels]] = []
+        metadata_list: list[Params] = []
+        for i in indices:
+            img_dict, labels, metadata = self._load_data(i)
+            loaded_anns.append((img_dict, labels))
+            metadata_list.append(metadata)
+
+        img_dict, labels = self.augmentations.apply(loaded_anns)
+        metadata = self._merge_record_metadata(metadata_list)
```
klemen1999: Nitpick: Let's use here "sample_metadata" for variable name and "_merge_sample_metadata" for function name
Elsewhere as well, let's be consistent and use "sample_metadata" for variable naming instead of just "metadata" for clearity

luxonis_ml/data/utils/parquet.py
```diff
@@ -15,6 +17,7 @@ class ParquetRecord(TypedDict):
     instance_id: int | None
     task_type: str | None
     annotation: str | None
+    sample_metadata: str
```
klemen1999, line 20: Is there a specific reason why we don't want to have this as `str | None`? I imagine then we wouldn't need to insert default metadata everywhere? Or is that trickier to implement than what is done currently?

luxonis_ml/data/__main__.py
```diff
@@ -222,6 +222,10 @@ def inspect(
         bool,
         Parameter(negative=""),
     ] = False,
+    show_metadata: Annotated[
```
klemen1999: Same nitpick: Let's use "show_sample_metadata" as variable name

luxonis_ml/data/__main__.py
```diff
@@ -298,11 +303,12 @@ def inspect(
...
-        if isinstance(img, dict):
-            images_dict = img
-        else:
-            images_dict = {"image": img}
+    for data in loader:
+        images_dict = data.images
+        labels = data.labels
+
+        if show_metadata:
+            print(data.metadata)
```
klemen1999: Nitpick: I'd maybe do `print("Sample metadata:", data.metadata)` so it's clearer what is being printed in the terminal

klemen1999 (commented):
- We probably need updates to the `luxonis_ml/docs/README.md` as well right? Possibly somewhere else as well?
- in `BaseLoader` and `LuxonisLoader` we are referencing `LuxonisLoaderOutput` type but is this correct? Should it be `LoaderOutput` since `LuxonisLoaderOutput` is not actually defined anywhere?
  - Is the idea to keep `BaseLoader` without `sample_metadata`? Could this cause issues with e.g. luxonis-train or some other package if they are inheriting from `BaseLoader` and the contract is the different ie. we need to handle cases with and without `sample_metadata`

luxonis_ml/data/utils/parquet.py
kozlov721: Not really a specific reason, we just already save it as empty detadata if not provided. I think we would need to do the same work if it could be None.

kozlov721:
> We probably need updates to the luxonis_ml/docs/README.md as well right? Possibly somewhere else as well?

I want to include this in the overall docs update

klemen1999 (approved):

PR #438, opened 2026-07-02
Added optional strides and background_class head metadata

klemen1999:
## Purpose

- Adds optional `strides` parameter to `HeadYOLOMetadata` in NNArchive
- Adds optional `background_class` to `HeadSegmentationMetadata` in NNArchive

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **New Features**
  * Added a `background_class` option for segmentation head metadata (defaulting to off).
  * Added optional per-head `strides` metadata for YOLO-based heads.

* **Bug Fixes**
  * Updated metadata validation so `strides` is treated as a shared/common parameter across validation phases.

* **Tests**
  * Extended archive config tests to assert `background_class` and `strides` are correctly exported.
  * Added serialization tests covering default and overridden head metadata values.


tests/test_nn_archive/heads.py
```diff
@@ -74,6 +76,7 @@
     anchors=None,
     n_prototypes=10,
     n_keypoints=None,
+    strides=[8, 16, 32],
```
dtronmans, line 79: Should we have some validation either here or on the `depthai-nodes` side that `len(strides)` == number of YOLO output heads? Because here in this test the strides is [8, 16, 32] but I also see `yolo_outputs=["feats"]` which are of different lengths, but in [this code snippet](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/node/parsers/utils/yolo.py#L525) I think we assume that there is one stride per output head.

dtronmans (commented):
Just one comment about number of output heads and number of strides being equivalent now that we actually set the strides

tests/test_nn_archive/heads.py
klemen1999: This is a good point and I slight adjusted the test [here](https://github.com/luxonis/luxonis-ml/pull/438/commits/7058a2bf053a21b48196055fd5b31dcd952980b9) to be semantically more correct. But I think the actual verification should happen in the depthai-nodes (or any consumer of the NNArchive) because this NNArchive Config is most of the time used more as a schema and not actual validator (mainly validates just that required fields are present and have correct types). I'll add this as a subtask to the existing deptahi-nodes "small bugfixes"

dtronmans (approved):

PR #440, opened 2026-07-08
CutMix Augmentation

kozlov721:
## Purpose

Increases our support for more advanced augmentations.

## Specification

Adds new `CutMix` batch augmentation to the registry.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: codex:gpt-5.5

## Summary by CodeRabbit

* **New Features**
  * Added CutMix as a batch augmentation for mixing image patches.
  * CutMix updates masks, bounding boxes, keypoints, metadata, and labels with visibility and occlusion handling.
  * Supports aspect-ratio-preserving or standard resizing for differently sized images.
* **Bug Fixes**
  * Improved validation messaging for incorrectly nested batch augmentations.
* **Documentation**
  * Added CutMix configuration details and examples.
* **Tests**
  * Added comprehensive CutMix coverage and expanded batched augmentation tests.


klemen1999:
I found some edge cases where seg mask seemed to be kept but the bbox is not? Any ideas where this is commit from?

klemen1999:
There is also a discussion to be had what to do with bounding boxes that overlap with new patch. From the model training perspective it would make the most sense that these would be clipped oterwhise you get issues like this one where the person on the "bottom layer" has its bounding box over the patch and from train perspective it doesn't make sense that the edge of the bbox would be there. This could lead to unstable training.



Edge case of clipping is the case where patch is in the middle of the bacground object. in that case I'm not sure what to do: try to clip witht he goal of perservice biggest possible area or maybe just drop the background bbox entirely. Could we do a bit of research into how this is handled by some other CutMix implementations? @kozlov721

kozlov721:
>I found some edge cases where seg mask seemed to be kept but the bbox is not? Any ideas where this is commit from?

This might be because the mask is a semantic mask which is unrelated to bboxes. The example COCO dataset has both semantic and instance masks. 

> There is also a discussion to be had what to do with bounding boxes that overlap with new patch. From the model training perspective it would make the most sense that these would be clipped oterwhise you get issues like this one where the person on the "bottom layer" has its bounding box over the patch and from train perspective it doesn't make sense that the edge of the bbox would be there. This could lead to unstable training.

Hm I could also see the opposite making sense from the training perspective. What I mean is that the person is supposed to occupy the entire bbox, just part of the person is momentarily covered. Then it's up for a debate whether it is more correct for the model to predict the "cropped" bbox or if the bbox should contain the occluded part of the object too. 

From what I could find the standard is to keep bboxes as they are (just hide them entirely if the object area falls under a specified threshold)

luxonis_ml/data/augmentations/albumentations_engine.py
```diff
@@ -138,6 +139,13 @@ class AlbumentationsEngine(AugmentationEngine, register_name="albumentations"):
     Batch Augmentations
     -----------------------------
 
+    `CutMix`
+    ~~~~~~~~
```
klemen1999, line 155: Nitpick: Line 156 (under Mosaic4) has 2 more `~` symbols than line 150 and line 143. Let's unify this

luxonis_ml/data/augmentations/custom/letterbox_resize.py
```diff
@@ -342,3 +342,15 @@ def _apply_to_image_data(
             padded_img = padded_img[..., None]
 
         return padded_img
+
+
+def create_letterbox_or_resize(
```
klemen1999, line 347: Nitpick: This feels more like a function that should be in `albumentations_engine` rather than `letterbox_resize`. What is the reason it was put here? Maybe circular import issue in mixup.py? Any idea of a better place for it? It feels a bit weird that `albumenations_engine` has to always import LetterboxResize

klemen1999 (approved):
I added a small fix for pathces with 0 size in https://github.com/luxonis/luxonis-ml/pull/440/commits/8931da1c46dc142a7fc8cc1bc5da0c4eca808264
Other than two small comments LGTM, ready to merge after merge conflicts are resolved

luxonis_ml/data/augmentations/albumentations_engine.py
kozlov721: The number of underline symbols must be the same as the number of letters in the title (that's the docformat spec).

klemen1999: Aaa ok, didn' know that, thanks

PR #458, opened 2026-07-28
Fix bbox ordering with multiple batch augmentations

kozlov721:
## Purpose

Fixes a bug triggered by the following MRE:

**`aug.json`**
```json
[
  {
    "name": "Mosaic4",
    "params": {
      "width": 640,
      "height": 640,
      "p": 0
    }
  },
  {
    "name": "MixUp",
    "params": {
      "p": 1
    }
  }
]
```

```
luxonis_ml data inspect coco_test -a aug.json
```

## Specification

None / not applicable

## Dependencies & Potential Impact

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Submitted code was reviewed by a human: YES/NO

The author is taking the responsibility for the contribution: YES/NO

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batched Albumentations augmentation to preserve per-annotation label metadata, keep empty/filtered bbox tasks, and maintain correct shapes and ordering for dependent labels (keypoints, masks, metadata).
  * Refined bbox-associated label compaction and index re-stamping across batch transforms.
  * Improved MixUp and Mosaic handling for empty/partially annotated inputs to keep batch dimensions stable.

* **Tests**
  * Expanded regression/property-based coverage for box↔keypoint/mask alignment, empty-box behavior, task filtering, and batch-compose warning/compaction semantics.

PR #462, opened 2026-07-30
De-AI pass over the batched-augmentation fix

kozlov721:
Readability pass over `fix/batched-augs`. **No behaviour change** — targeted at `fix/batched-augs` so the diff shows only this pass.

Net **−199 lines** (+217/−416). Tests: 195 passed, matching the branch baseline of 196 minus the one deleted test; the same 3 `test_loader.py` failures are pre-existing and caused by missing `GOOGLE_APPLICATION_CREDENTIALS`. `pre-commit run` and `pyright --warnings --level warning --pythonversion 3.10` are both clean.

## Source

- **`_compact_bbox_associated_labels`** had three near-identical branches, each with its own count check, its own 4-line `_warn_unmatched` call and its own `continue`. Extracted `_select_instances`, which returns `None` when a field cannot be matched to its boxes, leaving one warn call site. The warning now reports the field's shape instead of a pre-formatted English fragment (`"5 instances"` / `"5 values"` / `"5 keypoint rows"`) picked per target type.
- **`_get_task_group`** was an `endswith`/`else: return ""` branch; it is now a `removesuffix` pair. Checked equivalent against the old implementation over 4680 generated task strings, including malformed ones.
- The new bbox-grouping block had been **inserted between an existing comment and the `catch_warnings` statement it describes**; moved back together.
- Named `ASSOCIATED_TARGET_TYPES` instead of inlining the set literal, and renamed `bbox_target`/`bbox_type` (the loop walks *all* targets, not just bboxes).
- Dropped a comment in `mosaic.py` referring to "the guard above" 60 lines away.

## Tests

The strongest tell was documentation style. No test file on `main` has a module docstring (0/14) and about 1 in 18 tests has one at all; this branch added 7 module docstrings and ~50 test docstrings, many restating the test name. Removed those, kept the ones recording a real failure mode.

- One shared `KeepFirstSample` in `helpers.py` replaces three near-copies across three files.
- `test_compaction_handles_all_filtered_bboxes` looped over 10 seeds with an `all_filtered_seen` flag, `continue`, and a final assert that the case was ever reached — every seed in that range hits it, so it is now a single deterministic run.
- `test_task_groups_do_not_interfere` returned early whenever fewer than 2 groups were drawn, silently passing on most examples; it now requires 2 via `sample_specs(min_groups=2)`.
- `test_every_label_type_works_on_its_own` sampled a config *index* bounded by a magic `7` and re-imported `BATCH_CONFIGS` inside the test body while the same module was imported at the top; it now samples the list directly.
- Deleted `test_bbox_params_cannot_be_passed_positionally`, which used `cast(Any, ...)` to defeat the type checker in order to assert that Python rejects a positional keyword-only argument.
- Replaced `out_labels.get(key, np.zeros(...)).shape[0] == 0` with assertions on what the output actually contains.
- Removed `assert SomeClass.__name__ in TRANSFORMATIONS` lines that checked the fixture rather than the code under test.

## One judgement call worth review

I dropped `assert boxes.shape[0] <= group.n_instances * 8` from `test_instances_never_outnumber_their_bboxes`. The `8` was the max batch size, unrelated to the invariant the test is named for, and the remaining assertion covers that invariant. Happy to restore it with the bound derived from `engine.batch_size` if you'd rather keep the check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/luxonis_ml/data/augmentations/albumentations_engine.py
+++ b/luxonis_ml/data/augmentations/albumentations_engine.py
@@ -39,6 +39,12 @@ TargetType: TypeAlias = Literal[
     "metadata",
 ]
 
+# Target types describing individual instances, and so kept in step with the
+# bounding boxes of their task group as those are filtered.
+ASSOCIATED_TARGET_TYPES = frozenset(
+    {"instance_mask", "keypoints", "array", "metadata"}
+)
+
 
 class AlbumentationConfigItem(ConfigItem):
     """Configuration item for `AlbumentationsEngine`.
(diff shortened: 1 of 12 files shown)


kozlov721:
Opened by mistake — this work belongs on #458 as an additional commit, not as a separate PR.

PR #465, opened 2026-08-01
Stop filesystem tests leaking into the shared bucket

kozlov721:
## Purpose

`test_static_directory[s3]` failed on #464, a branch that changed nothing but a hashing routine ([failing job](https://github.com/luxonis/luxonis-ml/actions/runs/30713102692/job/91404209251)):

```
>       assert not fs.exists(dir_name)
E       AssertionError: assert not True
E        +  where True = exists('static_dir_upload_70182')
```

That assertion is a *precondition* — it checks the name is free before uploading — and it failed because the directory was already in the bucket.

The `s3://` and `gcs://` filesystems address a bucket shared by every run, on every platform and Python version. Five of the six tests that upload there never deleted what they created, so every run left more behind, permanently:

| test | remote artifact | cleaned up before |
| --- | --- | --- |
| `test_file` | `upload_file_{randint}.txt` | yes |
| `test_static_file` | `static_file_upload_{randint}.txt` | **no** |
| `test_directory` | `upload_dir_{randint}_{name,list}` | **no** |
| `test_static_directory` | `static_dir_upload_{randint}` | **no** |
| `test_walk_dir` | `dir_upload_dir_{randint}` | pre-guarded, never deleted |
| `test_bytes` | `bytes_test_{randint}.txt` | **no** |

Four of those open by asserting the name is free, and the names come from `randint`, drawn from a range of 100_000. So the collision probability grows with every run — this was not a one-off, it gets steadily more likely.

## Specification

Adds a `remote` fixture that reserves names and removes them when the test finishes, and routes every remote upload in the module through it. It follows the idiom already used by `dataset_name` in `tests/conftest.py`, which yields and then deletes remotely.

Reserving a name **clears** it as well as scheduling it for deletion. That part is what makes the fix self-correcting rather than merely forward-looking: the bucket still holds every leftover from before this change, and a run killed between its upload and its teardown will keep adding more. Clearing on the way in means neither can fail the next test that draws the same name.

This also removes the two hand-rolled `if fs.exists(...): delete` pre-guards in `test_file` and `test_walk_dir`, which were doing half of this job in two places.

Tests only — no library code is touched.

The one behavioral consequence worth naming: the suite now issues extra `exists`/`delete` calls against the shared bucket, one pair per reserved name. That is the point of the change, but it does mean slightly more remote traffic per run.

Existing leftovers in the bucket are **not** purged by this PR — that needs credentials and is a one-off cleanup. It is no longer urgent, though, since a collision with an old leftover is now cleared rather than fatal.

## Deployment Plan

None / not applicable — test-only change, nothing to roll out.

## Testing & Validation

Adds `test_remote_names_are_cleared_before_use`, which plants a leftover file and a leftover directory and asserts that reserving the name frees it. Both kinds are covered because they are removed through different calls (`delete_file` vs `delete_dir`). It reserves each name *before* planting the leftover, so a failing assertion cannot leak the very thing the test is about.

## AI Usage

Assisted-by: Claude Code:claude-opus-5 [pyright] [ruff]

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_utils/test_filesystem.py
+++ b/tests/test_utils/test_filesystem.py
@@ -1,6 +1,6 @@
 from collections.abc import Callable, Iterator
-from contextlib import suppress
 from pathlib import Path
+from uuid import uuid4
 
 import pytest
 from _pytest.fixtures import SubRequest
@@ -73,13 +73,12 @@ def fs(
 
 def _remove_remote(fs: LuxonisFileSystem, name: str) -> None:
     """Delete a remote file or directory, if it is there at all."""
-    with suppress(Exception):
-        if not fs.exists(name):
-            return
-        if fs.is_directory(name):
-            fs.delete_dir(name)
-        else:
-            fs.delete_file(name)
+    if not fs.exists(name):
+        return
+    if fs.is_directory(name):
+        fs.delete_dir(name)
+    else:
+        fs.delete_file(name)
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


klemen1999 (approved):
LGTM, I just added a small addition to fix concurant runs and fail on delete exceptions to prevent leaks.

PR #466, opened 2026-08-01
Shared color utilities in `luxonis_ml.utils.color`

kozlov721:
## Purpose

There was no first-class way to handle colors in the ecosystem. The only color parser available was matplotlib's, which returns floats in `[0, 1]` and drags a plotting dependency into code that just wants an RGB triple (see the letterbox bug in the follow-up PR). Anything that draws — augmentations, dataset health charts, visualization — needs one shared, dependency-free color model.

Split out of the `feat/vizlab` branch, which had grown to 81 commits; this is one of eight self-contained pieces that do not belong to vizlab itself.

## Specification

New `luxonis_ml.utils.color` package, re-exported as `luxonis_ml.utils.Color`:
- `base.py` — the `Color` model and `ColorLike` alias: parsing from names, hex, grayscale ints and RGB(A) tuples, with 0-255 channels and clamping.
- `brand.py` — the Luxonis brand colors and background-aware chrome.
- `gradient.py` — named gradients and interpolation.
- `palette.py` — stable, visually distinct per-label color assignment.

Purely additive: no existing module changes behavior in this PR.

## Dependencies & Potential Impact

No new third-party dependencies — the package is pure Python and stdlib only. Additive API, so nothing breaks. Later PRs in this series build on it.

## Deployment Plan

None / not applicable

## Testing & Validation

`tests/test_utils/test_color.py` — 14 tests covering parsing, clamping, conversion, gradients and palette stability. `ruff`, `pyright` and the pydoctor build are clean.

## AI Usage

Assisted-by: Claude:claude-fable-5

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive color utilities for parsing, validation, transformations, and readable text color selection.
  * Added brand colors, light/dark visualization palettes, semantic accents, and adaptive display styling.
  * Added customizable palettes with deterministic generation, pinned colors, and class-based assignment.
  * Added gradient generation, interpolation, preset gradients, and field colorization support.

PR #467, opened 2026-08-01
Fix letterbox padding with named and hex fill colors

kozlov721:
## Purpose

`LetterboxResize` resolved its fill colors through `luxonis_ml.data.utils.visualizations.resolve_color`, which delegated string colors to `matplotlib.colors.to_rgb` — that returns floats in `[0, 1]`. So `image_fill_value="white"` padded images with `(1, 1, 1)` on a uint8 image, i.e. effectively black, silently producing the opposite of what was asked for.

Split out of the `feat/vizlab` branch, which had grown to 81 commits; this is one of eight self-contained pieces that do not belong to vizlab itself.

## Specification

Resolve fill values through `luxonis_ml.utils.color.Color`, which yields proper 0-255 channels, via a small `_fill_color()` helper.

`Color.parse` clamps out-of-range channels, whereas `resolve_color` raised. A fill value outside `[0, 255]` is a config mistake rather than something to round off, so `_fill_color` keeps rejecting it — and now names the offending parameter in the message. `mask_fill_value` goes through the same path.

Behavior change, and an intentional one: a named or hex `image_fill_value` now pads with the requested color instead of black. Anyone who compensated for the old behavior by passing an explicit tuple is unaffected. Also removes the last non-vizlab consumer of `data.utils.visualizations`, which the vizlab work deletes.

## Deployment Plan

None / not applicable

## Testing & Validation

`tests/test_data/test_augmentations/test_letterbox.py` — 4 new tests (9 total in the file, all passing): named/tuple/default fills resolve to the right RGB, out-of-range values are rejected for both fill parameters, and the full valid range is accepted. The first is a direct regression test for the bug above.

## AI Usage

Assisted-by: Claude:claude-fable-5

---

> **Stacked on #466** (`feat/color-utils`). Based on that branch so the diff shows only this feature; GitHub retargets it to `main` once #466 merges.

### DIFF
--- a/luxonis_ml/data/augmentations/custom/letterbox_resize.py
+++ b/luxonis_ml/data/augmentations/custom/letterbox_resize.py
@@ -6,35 +6,7 @@ import numpy as np
 from typing_extensions import override
 
 from luxonis_ml.typing import RGB
-from luxonis_ml.typing import Color as ColorLike
-from luxonis_ml.utils.color import Color
-
-
-def _fill_color(value: ColorLike, name: str) -> RGB:
-    """Resolve a user-supplied padding value to an ``(r, g, b)`` triple.
-
-    `Color.parse` clamps channels into ``[0, 255]``. A fill value outside that
-    range is a config mistake rather than something to round off, so it is
-    rejected here instead of silently padding with a different color.
-
-    Args:
-        value: A color name, hex string, grayscale int, or ``(r, g, b)`` tuple.
-        name: The parameter name, used in the error message.
-
-    Returns:
-        The resolved RGB triple.
-
-    Raises:
-        ValueError: If any channel falls outside ``[0, 255]``.
-
-    """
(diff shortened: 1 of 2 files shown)


luxonis_ml/data/augmentations/custom/letterbox_resize.py
```diff
@@ -5,8 +5,36 @@
...
 from typing_extensions import override
 
-from luxonis_ml.data.utils.visualizations import resolve_color
-from luxonis_ml.typing import RGB, Color
+from luxonis_ml.typing import RGB
+from luxonis_ml.typing import Color as ColorLike
+from luxonis_ml.utils.color import Color
+
+
+def _fill_color(value: ColorLike, name: str) -> RGB:
```
klemen1999: Nitpick: I'd prefer to have this function under class definition

klemen1999 (approved):
LGTM

PR #470, opened 2026-08-01
Rebuild dataset records from loader output

kozlov721:
## Purpose

`LuxonisLoader` returns flat per-task arrays optimized for training, while visualization and inspection code works with canonical `DatasetRecord` and `Detection` objects. This adds a supported conversion back to LDF records without requiring consumers to reconstruct instance relationships themselves.

Split out of the `feat/vizlab` branch, which had grown to 81 commits; this is one of eight self-contained pieces that do not belong to vizlab itself.

## Specification

`LoaderOutput` now carries the loader's images, labels, record metadata, class mapping, and categorical metadata encodings. `LoaderOutput.to_ldf()` rebuilds the label arrays into one `DatasetRecord` per task name:

```python
sample = loader[0]
records = sample.to_ldf()
```

The conversion supports classification, bounding boxes, keypoints, semantic and instance segmentation, array annotations, nested task names, and per-instance metadata. Class and categorical mappings attached by `LuxonisLoader` are used by default and can be overridden explicitly. Semantic-segmentation background masks remain hidden by default and can be retained with `render_background=True`.

The loader now keeps spatial annotations, class IDs, and metadata aligned by instance ID. Missing per-instance metadata is padded with `None`, and an entirely absent metadata task produces an empty array instead of phantom box-less detections.

Reconstructed records hold images and array annotations in memory. They can be rendered and inspected, but adding or serializing them is rejected until those arrays are persisted to files.

The `LoaderOutput.to_ldf()` API is additive, and the existing two-value `LoaderOutput` tuple protocol remains available for backward-compatible unpacking.

`DatasetRecord.files` and `ArrayAnnotation` now accept in-memory arrays for reconstructed records. Dataset ingestion still requires file-backed media and array annotations and raises a clear `NotImplementedError` otherwise.

The empty-label shape for an absent metadata task changes from `(n_classes,)` to `(0,)`. Optional metadata rows are aligned to spatial instances and use `None` for instances that do not contain the field.

## Deployment Plan

None / not applicable

## Testing & Validation

`tests/test_data/test_label_converter.py` contains 15 tests covering task reconstruction, nested task grouping, class and categorical decoding, optional metadata alignment, background handling, attached images and arrays, and persistence rejection.

Focused validation also covers loader ordering, absent metadata behavior, LDF in-memory validation, and static typing. The affected test set passes with 63 tests, and Pyright reports no errors.

## AI Usage

Assisted-by: Claude:claude-fable-5

---

> **Stacked on #469** (`feat/record-multiple-annotations`). Based on that branch so the diff shows only this feature; GitHub retargets it to `main` once #469 merges.


luxonis_ml/data/loaders/__init__.py
```diff
@@ -132,6 +132,37 @@
...
      - Original value structure.
      - Values keyed by metadata field name.
+   * - ``array``
+     - :math:`\left(N, C, \ldots\right)`
+     - Whatever each ``.npy`` held, wrapped in an instance and a class axis.
+
+Array labels
+------------
+
+``array`` is the escape hatch, and the only layout that is not self-describing:
```
klemen1999: This whole section feels a bit over AI generated. I'd adjust the wording here to what we actually want to say

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -404,12 +404,16 @@ def _add_empty_annotations(
...
                         )
-                    elif task_type == "classification" or task_is_metadata(
-                        task
-                    ):
+                    elif task_type == "classification":
                         labels[task] = np.zeros(
                             (len(self._classes[task_name]),)
                         )
+                    elif task_is_metadata(task):
+                        # Metadata is per-instance, so an absent metadata task is
```
klemen1999: Is this comment needed? Or can it be more to the point?

luxonis_ml/data/loaders/label_converter.py
```diff
@@ -0,0 +1,285 @@
...
+    """
+    per_instance: dict[int, dict[str, int | float | str | Category]] = (
+        defaultdict(dict)
+    )
+    for task_type, array in task_types.items():
+        if not task_type.startswith("metadata/"):
+            continue
+        key = task_type[len("metadata/") :]
+        values = np.asarray(array)
+        if values.ndim != 1 or len(values) != n_instances:
```
klemen1999: Optional per-instance metadata is silently dropped. The loader emits only values for annotations that contain a metadata key, so two boxes with metadata on only one yield a metadata array of length 1 vs. 2 instances. The length check skips it entirely. Preserve instance alignment in loader output (or explicitly reject this unsupported case); otherwise valid labels disappear in reconstructed records.

And a practical edge case:
A detection may omit a metadata field while another detection in the same task supplies it. For example:
Original records
```
car:    bbox + {"tag": "kept"}
person: bbox
```
The loader produces:
```
"det/boundingbox":  # 2 rows
"det/metadata/tag": ["kept"]  # 1 row
```
The converter calculates n_instances == 2, then requires metadata to have exactly two elements. Since it has one, it discards the whole field.
That means the reconstructed result is:
```
car:    bbox + {}
person: bbox + {}
```
The converter cannot safely infer whether "kept" belonged to the first or second box. The loader format needs either an instance-ID array for metadata or a placeholder value per missing instance.

luxonis_ml/data/loaders/label_converter.py
```diff
@@ -0,0 +1,285 @@
...
+                )
+            )
+        detections.append(
+            Detection(
+                instance_id=i,
+                class_name=class_name,
+                boundingbox=boundingbox,
+                keypoints=keypoint_annotation,
+                instance_segmentation=instance_segmentation,
+                metadata=metadata.get(i, {}),
```
klemen1999: Metadata is paired by row index, but the loader does not guarantee the same ordering: spatial annotations are sorted by instance_id, while metadata remains in source-row order. With records emitted as instance 1 then instance 0, the converter attaches each value to the other box. The loader needs to sort/pad metadata alongside annotations or expose instance IDs.

Practical example:
Even if every instance has metadata, the loader does not keep its order aligned with spatial labels.
Given source annotations emitted in this order:
```
instance_id=1: person, tag="one"
instance_id=0: car,    tag="zero"
```
The loader sorts boxes/keypoints/masks by instance_id, but metadata is collected in source order. So the resulting arrays are effectively:
```
boxes:    [instance 0 geometry, instance 1 geometry]
metadata: ["one", "zero"]  # still source order: 1, then 0
```
The converter pairs by row index, resulting in instance 0 -> "one" and instance 1 -> "zero"—reversed.
There is a related loader issue: its class-ID list is not sorted alongside the spatial annotations, so out-of-order source records can also attach the wrong class to the sorted box geometry before the converter sees it. The converter exposes these existing ordering problems because it assumes all arrays are row-aligned.

luxonis_ml/data/loaders/label_converter.py
```diff
@@ -0,0 +1,285 @@
...
+        decoder = None
+        if full_task in categorical_encodings:
+            decoder = {
+                index: name
+                for name, index in categorical_encodings[full_task].items()
+            }
+        for i, value in enumerate(values):
+            item = value.item() if hasattr(value, "item") else value
+            if decoder is not None:
+                item = decoder.get(int(item), item)
```
klemen1999: Passing valid output from LuxonisLoader(keep_categorical_as_strings=True) together with categorical_encodings raises ValueError on int("sunny"). Decode only numeric encoded values; leave already-string categorical values unchanged.

Practical example:
By default, categorical metadata is numeric in loader output:
```
["sunny"] -> [0]
```
When categorical encodings are supplied, the converter correctly turns 0 back into "sunny".
But LuxonisLoader(keep_categorical_as_strings=True) is explicitly supported and returns:
```
["sunny"]
```
With the same encoding map supplied, the converter does int("sunny"), raising ValueError. It should only decode numeric values and pass strings through unchanged.

klemen1999 (changes requested):
Some issues found by AI assisted review

luxonis_ml/data/loaders/label_converter.py
```diff
@@ -0,0 +1,292 @@
...
+    )
+    detections.extend(
+        _semantic_detections(task_types, id_to_name, render_background)
+    )
+    # Bounding boxes and semantic masks carry their classes directly, so their
+    # derived classification vector would be redundant. Keypoints and instance
+    # masks do not encode class ids, however, so retain their vector as
+    # record-level class tags when none of the reconstructed instances is named.
+    if not any(detection.class_name is not None for detection in detections):
+        detections.extend(_classification_detections(task_types, id_to_name))
```
klemen1999: AI found issue:
The LDF model permits an annotation list such as:
```
DatasetRecord(
    file="image.jpg",
    task_name="det",
    annotation=[
        Detection(
            class_name="car",
            boundingbox={"x": 0.1, "y": 0.1, "w": 0.2, "h": 0.2},
        ),
        Detection(class_name="scene_tag"),
    ],
)
```
`Detection(class_name="scene_tag")` is a valid image-level classification annotation; classification is represented by `class_name` without a spatial field. The loader produces both:
```
"det/boundingbox": [[car_id, ...]]
"det/classification": [1, 1]  # car and scene_tag
```
The converter sees the named `car` box and therefore skips the whole classification vector. That loses `scene_tag`. The comment is correct for classification bits already implied by boxes/masks, but it needs to subtract those represented classes and emit the remaining active classes.

luxonis_ml/data/loaders/label_converter.py
```diff
@@ -0,0 +1,292 @@
...
+    Returns:
+        One `DatasetRecord` per task name.
+
+    """
+    grouped: dict[str, dict[str, np.ndarray]] = defaultdict(dict)
+    for key, array in labels.items():
+        task_name, task_type = _split_loader_key(key)
+        grouped[task_name][task_type] = array
+
+    files: dict[str, np.ndarray | Path] = (
```
klemen1999: AI found issue:
All returned `DatasetRecord`s share the same mutable `files` dictionary. With multiple task records, mutating `records["detection"].files` also changes `records["segmentation"].files`. Construct a fresh mapping per record (`files=dict(files)`) to keep records independent.

Suggested fix:
Move the `files` copy into the per-task loop, so each `DatasetRecord` gets its own dictionary:

```python
files: dict[str, np.ndarray | Path] = (
    dict(images) if images else {"image": Path("<loader>")}
)

records = {}
for task_name, task_types in grouped.items():
    records[task_name] = DatasetRecord.model_construct(
        files=dict(files),  # independent mapping per record
        annotation=detections,
        task_name=task_name,
        sample_metadata={},
    )
```

This is a shallow copy: each record still references the same image arrays, which is desirable and avoids copying image data. Only the `files` mapping becomes independent, so changing one record’s sources does not affect the others.

PR #476, opened 2026-08-02
Migration to `uv`

kozlov721:
## Purpose

Move dependency and build management to `uv`, so local development, CI and publishing all
resolve from one locked environment instead of the nine hand-maintained requirement files
scattered across the repo.

This mirrors luxonis/luxonis-train#421 — **see that PR for the full rationale** behind `uv`,
`prek`, the generated requirements exports and the `opencv-python-headless` override. Only what
is specific to `luxonis-ml` is written out below.

## Specification

- Dependencies now live in `pyproject.toml`: runtime in `[project.dependencies]`, the extras in
  `[project.optional-dependencies]`, tooling in the `lint` / `dev` / `docs` PEP 735 groups. The
  four per-module `requirements.txt` files and `extra_requirements/` are gone.
- Added `.python-version` and `uv.lock`.
- `requirements.txt`, `requirements-all.txt` and `requirements-dev.txt` are now generated pip
  compatibility exports produced by `scripts/export_requirements.sh` and refreshed by a
  pre-commit hook. Don't edit them by hand.
- CI, publishing, docs and `CONTRIBUTING.md` use `uv sync`, `uv run` and `uv build`.
- `pre-commit` → [`prek`](https://github.com/j178/prek). Same `.pre-commit-config.yaml` and the
  same upstream hook repos, so it is a drop-in swap; it also drops the `pre-commit<4.0.0` cap.

### Development commands

| Before | Now |
|---|---|
| `pip install -e '.[dev]'` | `uv sync` |
| `pre-commit install` | `uv run prek install` |
| `pre-commit run --all-files` | `uv run prek run --all-files` |
| `pytest tests --only-local -n auto` | `uv run pytest tests --only-local -n auto` |
| `pyright --project pyproject.toml` | `uv run pyright --warnings --level warning --project pyproject.toml` |
| `python tools/build_pydoctor_docs.py …` | `uv run --only-group docs python tools/build_pydoctor_docs.py …` |
| editing `requirements*.txt` by hand | edit `pyproject.toml`, then `scripts/export_requirements.sh` |

`uv sync` installs the project with **all** extras plus the `dev` group, which is what `.[dev]`
used to mean. No `source .venv/bin/activate` step — `uv run` syncs and runs in `.venv`.

### Fixed along the way

- **`[tool.pyright] exclude` was replacing pyright's defaults.** Harmless until now, but once
  `uv sync` puts a `.venv` in the project root, pyright walks into it and tries to check every
  installed dependency — it hangs indefinitely. The default patterns are restored, otherwise the
  `type-check` job would have hung in CI. It now checks 180 files in ~13 s.
- **`packages.find` had no `include` filter**, so the published wheel shipped `tests` and `tools`
  as top-level packages. Scoped to `luxonis_ml*`.
- `lint` is a group separate from `dev` so the hooks job installs a single binary instead of the
  whole tree that `dev` pulls in through `luxonis-ml[all]`.
- Dropped two now-dead pip workarounds: the `PIP_CONSTRAINT=setuptools<81` dance, and
  `pip install 'pillow-heif<0.22.0'` (immediately upgraded away by `pillow-heif~=1.5`).

- **`pip install -e '.[dev]'` no longer works.** `dev` is a group now, not a published extra, and
  pip treats an unknown extra as a warning — so that command exits 0 having installed no dev
  tooling. Use `uv sync`. External scripts using the old invocation need updating.
- `requirements-dev.txt` changed meaning: it was a 10-line additive overlay, and is now a
  standalone full export of base + all extras + `dev`.
- `opencv-python-headless` is no longer installed.
- The published wheel no longer contains the `tests` and `tools` packages.
- Contributors have to re-install the git hook once after pulling — `uv run prek install -f` —
  since the existing `.git/hooks/pre-commit` shim points at `pre-commit`.
- `requires-python` stays `>=3.10` (not narrowed to `<3.13` as in the luxonis-train PR); the lock
  resolves cleanly across the whole range.

## Deployment Plan

None / not applicable — packaging and CI only, no runtime behavior changes.

## Testing & Validation

- `prek run --all-files` green under the exact CI invocation; `export-requirements` is
  idempotent (a second run produces no diff).
- `uv lock --check` clean. `uv build` produces both artifacts, and the wheel contains only
  `luxonis_ml`.
- Pyright: 180 files, 0 errors, 0 warnings.
- 153 passed across `test_utils` / `test_nn_archive` / `test_telemetry` / `test_cli` /
  `test_typing`, and 238 passed in `tests/test_data`. Every remaining local failure is a Google
  `RefreshError: Reauthentication is needed` from absent cloud credentials — there are no
  assertion, import or dependency errors anywhere in the run.
- Reproduced the `check-requirements` job for `data`, `nn_archive` and `utils`: each extra syncs
  and imports on its own. Docs build verified through `--only-group docs`.

## AI Usage

Assisted-by: Claude:claude-opus-5

## Summary by CodeRabbit

- **Improvements**
  - Standardized setup and CI execution on reproducible, locked environments using pinned dependency installs.
  - Added/updated autogenerated, fully pinned requirement exports (including environment-specific markers) for consistent installation across Python versions and platforms.

- **Documentation**
  - Updated installation and contributor guides to use the new environment workflow and requirement export process.

- **Automation**
  - Updated CI and publish pipelines to use the locked workflow.
  - Added scheduled dependency auto-update and an installation verification workflow to ensure exported requirements stay current.


tools/export_requirements.sh
dtronmans, line 1: This is a bit more of a stylistic comment, but we already have the `tools/` subdirectory in this repository which I guess is for one-off scripts. Could this be moved to `tools/` instead, just to not add another subdirectory? However as I said this is more of a stylistic choice so either way is fine

dtronmans (approved):
Except for one stylistic comment I think everything looks good, I tried it locally through `uv sync` and everything works fine

klemen1999 (approved):

PR #483, opened 2026-08-11
Native export targeting older LDF versions

kozlov721:
## Purpose
`DatasetRecord` forbids extra fields, so a record carrying a field the reading installation does not know fails validation outright. LDF 2.1 added `sample_metadata`, which made every native export unreadable by an older luxonis-ml, with no way to ask for one it could read and no way to tell afterwards which version an export directory held.

`luxonis_ml data export ds --type native --ldf-version 2.0` now writes an export an older install parses cleanly.

## Specification

Records are built at the current `LDF_VERSION` and passed through an ordered chain of per-step downgrade transforms on the way out -- the counterpart of `migration.py`, which upgrades a stored dataframe on load. Today the chain holds one step, 2.1 -> 2.0, dropping `sample_metadata`. Supporting a future version is one prepended entry, and a test fails if `LDF_VERSION` is raised without a matching step.

Downgrading is lossy, so it warns with the dropped field and the record count, staying quiet when nothing populated was actually discarded.

Every export root (and every partition) now carries a `metadata.json` naming the version written. It is only a version stamp, deliberately not the full `Metadata` model a dataset keeps in its own storage. `NativeParser` reads it solely to warn about exports from a *newer* version; an older record is a strict subset of what it accepts, so reading one needs no transform, and a missing stamp stays non-fatal for existing exports.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

## Summary by CodeRabbit

* **New Features**
  * Native dataset exports can target a specific LDF version, including supported older versions.
  * Exported partitions now include metadata identifying their LDF version.
  * Older-version exports automatically remove unsupported fields and report discarded data.

* **Bug Fixes**
  * Added validation for unsupported, malformed, or incompatible LDF versions.
  * Added compatibility warnings when loading exports created with newer LDF versions.
  * Prevented LDF version selection for non-native export formats.

* **Documentation**
  * Clarified LDF version constraints, downgrade behavior, and metadata handling.

### DIFF
--- a/luxonis_ml/data/parsers/native_parser.py
+++ b/luxonis_ml/data/parsers/native_parser.py
@@ -3,7 +3,11 @@ from contextlib import suppress
 from pathlib import Path
 from typing import Any
 
+from loguru import logger
+from semver.version import Version
+
 from luxonis_ml.data import DatasetIterator
+from luxonis_ml.data.utils.constants import LDF_VERSION
 from luxonis_ml.typing import PathType
 from luxonis_ml.utils.path import resolve_manifest_path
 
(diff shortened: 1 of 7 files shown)


luxonis_ml/data/parsers/native_parser.py
```diff
@@ -40,17 +44,42 @@ def _resolve_annotation_paths(
...
+    only a newer one is worth reporting. The stamp is best-effort: exports
+    predating it have none, and a damaged one must not break the parse.
+    """
+    stamp = annotation_path.parent.parent / "metadata.json"
+    with suppress(OSError, ValueError, TypeError, KeyError):
+        version = Version.parse(
+            json.loads(stamp.read_text())["ldf_version"],
+            optional_minor_and_patch=True,
+        )
+        if version.major > LDF_VERSION.major:
```
klemen1999: Should we here rather compare `version` to `LDF_VERSION` instead of `.major` versions of them? Since for example we added last addition between 2.0 and 2.1

klemen1999 (commented):
Smaller comments, generally LGTM

klemen1999 (approved):

PR #488, opened 2026-08-14
Remove the submodule READMEs and build the API reference from docstrings

kozlov721:
## Purpose

The submodule `README.md` files held a second copy of the API documentation.
They drifted from the code, so readers had two sources of truth. This PR
removes them. The docstrings become the only home for API documentation, and
the generated reference comes from the docstrings alone.

## Specification

- Removes the seven submodule `README.md` files. Their content moves into the
  package docstrings.
- Rewrites the root `README.md`. It follows the layout of the other Luxonis
  repositories, and it links to the documentation portal.
- Removes `tools/build_pydoctor_docs.py`. The CI job runs `pydoctor luxonis_ml`
  instead.
- Adds a `[tool.pydoctor]` section to `pyproject.toml`. pydoctor reads the
  Google docstring format from there.
- Removes the `docs-pages.yaml` workflow. GitHub Pages no longer hosts the
  reference.
- Updates `CONTRIBUTING.md` for the new documentation workflow.

The published reference moves from GitHub Pages to the Luxonis documentation
portal. The `Export pydoctor reference` workflow builds it. The old GitHub
Pages URLs no longer receive updates.

The Python API does not change. This PR touches only docstrings, Markdown, and
CI configuration.

## Deployment Plan

None / not applicable

## Testing & Validation

- `pre-commit run --all-files`: all 18 hooks pass.
- `pydoctor luxonis_ml`, the CI command: exits 0, and reports no docstring
  syntax errors. The warning count stays at 28, so this PR adds none.
- `pytest --collect-only`: 984 tests collect.
- The package doctests: 57 pass.

Checks of the documentation against the code:

- All 10 documentation portal links return HTTP 200. A bogus path returns 404,
  so the check is meaningful.
- Every CLI command and subcommand in the `README.md` exists.
- The `" Class"` column name and the `random_split` default match the code.

These checks found four errors, which this PR corrects:

- The `README.md` used `--type ultralytics-ndjson`, which the CLI rejects. The
  correct value is `ultralyticsndjson`.
- The `README.md` listed a `sync` command, which does not exist. The commands
  are `push` and `pull`.
- `CONTRIBUTING.md` said that pydoctor exits 0 for a malformed docstring.
  pydoctor exits non-zero, so CI does fail on broken markup.
- `CONTRIBUTING.md` said that the export workflow hands the result to
  `luxonis/docs-content`. The workflow uploads an artifact instead.

## AI Usage

Assisted-by: Claude:claude-opus-5 pydoctor ruff pre-commit

## Summary by CodeRabbit

* **Documentation**
  * Reworked the README with expanded installation, quick-start, usage, CLI, telemetry, tracking, and contribution guidance.
  * Updated API documentation links, build instructions, and documentation portal guidance.
  * Added clearer guidance for datasets, splits, parsers, augmentations, telemetry, tracking, annotations, and optional integrations.
  * Removed outdated component-specific documentation and clarified dataset, parsing, telemetry, tracker, and annotation behavior.
  * Added deprecation notices for legacy dataset source classes.

* **Chores**
  * Simplified documentation checks and removed the previous automated documentation deployment workflow.


README.md
```diff
@@ -1,106 +1,377 @@
...
 
-Installing the package with the `data` extra and dependencies for `gcs` and `roboflow`:
+## 🧩 Modules
+
+Each module links to its own API reference:
+
+| Module                                                                                                                                            | Extra        | Purpose                                                                    |
+| ------------------------------------------------------------------------------------------------------------------------------------------------- | ------------ | -------------------------------------------------------------------------- |
+| [`luxonis_ml.data`](https://docs.luxonis.com/software-v3/ai-inference/model-source/training/luxonis-ml/luxonis-ml-api-reference/data)             | `data`       | Dataset creation, conversion, loading, augmentation, and export.           |
+| [`luxonis_ml.ldf`](https://github.com/luxonis/luxonis-ml/tree/main/luxonis_ml/ldf)                                                                | `ldf`        | The annotation schemas of the Luxonis Data Format. A subset of `data`.     |
```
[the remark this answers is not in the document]
kozlov721: Needs new online docs first

luxonis_ml/data/parsers/__init__.py
```diff
@@ -66,67 +66,75 @@
      - Dataset type
      - Parser
      - Typical annotations
-   * - COCO JSON
+   * - `COCO JSON <https://roboflow.com/formats/coco-json>`__
```
klemen1999: Should we link also to the Fiftyone format here since we support both?

luxonis_ml/data/README.md
```diff
@@ -1,1134 +0,0 @@
-> [!IMPORTANT]
```
klemen1999, line 1: One thing that we need to make sure is that if we remove this all of the information written must still be nicely present in the "new docs". This means that for example this README needs to be integrated into the `data.__init__.py` "fully". For example things like `generator()` I don't see documented now. Or some other similar things.
Check out info that is now in "standalone" docs in [docs.luxonis](https://docs.luxonis.com/software-v3/ai-inference/model-source/training/luxonis-ml/) from user PoV and test if all of the main information that is written there can now be easily navigated to in these new docs. At the end we'll probably still need some "standalone" docs for really common paths but the API docs should still include all of the common information easy to navigate to.

luxonis_ml/telemetry/README.md
```diff
@@ -1,356 +0,0 @@
-# LuxonisML Telemetry
```
klemen1999, line 1: I feel like a lot of general information is lost going from README -> `__init__.py`. In my mind I would have top level `__init__.py` be almost a copy-paste of current README since that gives the user main info + some useful examples how to use the module

luxonis_ml/tracker/README.md
```diff
@@ -1,168 +0,0 @@
-# LuxonisML Tracker
```
klemen1999, line 1: Similar comment as with telemetry, I would like to keep more (if not all) info now that we move from README -> `__init__.py`

README.md
```diff
@@ -1,106 +1,377 @@
...
-- `luxonis-ml[all]`: installs all dependencies
+1. **Convert a dataset into LDF**
 
-For a development environment, see [CONTRIBUTING.md](CONTRIBUTING.md); the development
-tooling lives in `uv` dependency groups rather than in a published extra.
+   We will use a sample COCO dataset from `RoboFlow` in this example.
 
-### Additional dependencies
+   ```bash
+   luxonis_ml data parse "roboflow://team-roboflow/coco-128/2/coco" --name coco_test
```
klemen1999, line 55: This internally calls `_download_roboflow_dataset()` which raises if `ROBOFLOW_API_KEY` is not set. I'm not sure if public datasets can be downloaded without setting the key. If they can be then it might be worth adjusting the `_download_roboflow_dataset()` to try first without key and then raise. If key is needed also for public datasets then it might be worth adding this as a note here in the QuickStart

README.md
```diff
@@ -1,106 +1,377 @@
...
+
+
+dataset.add(records())
+dataset.make_splits({"train": 0.8, "val": 0.1, "test": 0.1})
+```
+
+LDF also supports keypoints, semantic and instance segmentation,
+classification, arrays, and free-form metadata. See
+[`luxonis_ml.data.datasets`](https://docs.luxonis.com/software-v3/ai-inference/model-source/training/luxonis-ml/luxonis-ml-api-reference/data/datasets)
+for the dataset contract, and the [`luxonis_ml.ldf`](TODO) module for every annotation
```
klemen1999: We should be able to "guess"/predict the link here in TODO. Most likely it will be `https://docs.luxonis.com/software-v3/ai-inference/model-source/training/luxonis-ml/luxonis-ml-api-reference/ldf`

README.md
```diff
@@ -1,106 +1,377 @@
...
 
-The `luxonis-ml` package comes with a CLI that can be used to interact with the library.
+**Dataset storage is configured with:**
 
-To see the available commands, run:
+- `LUXONISML_BASE_PATH` - local base path for datasets and cache files, `~/luxonis_ml` by default
+- `LUXONISML_TEAM_ID` - team identifier used by dataset storage, `offline` by default
+- `LUXONISML_BUCKET` - the cloud bucket that holds remote datasets
+
+> [!NOTE]
```
klemen1999, line 354: This is not currently (yet) true. We don't collect telemetry on LuxonisML, we just have the module implemented here.

README.md
kozlov721: Oh that's right. I took it from lux-train where we're using it the same way so it would be good to update the README there too. We do mention the required key  both here and there but only at the very end and nothing makes it clear that it needs to be set for the README code to work.

luxonis_ml/data/README.md
kozlov721: Yeah that was the idea but something got lost anyway I see, thanks for spotting it.

README.md
kozlov721: [PR](https://github.com/luxonis/luxonis-train/pull/438/changes) for lux-train opened too

README.md
```diff
@@ -1,106 +1,384 @@
...
+<a name="modules"></a>
+
+## 🧩 Modules
+
+Each module links to its own API reference:
+
+| Module                                                                                                                                            | Extra        | Purpose                                                                    |
+| ------------------------------------------------------------------------------------------------------------------------------------------------- | ------------ | -------------------------------------------------------------------------- |
+| [`luxonis_ml.data`](https://docs.luxonis.com/software-v3/ai-inference/model-source/training/luxonis-ml/luxonis-ml-api-reference/data)             | `data`       | Dataset creation, conversion, loading, augmentation, and export.           |
+| [`luxonis_ml.ldf`](https://github.com/luxonis/luxonis-ml/tree/main/luxonis_ml/ldf)                                                                | `ldf`        | The annotation schemas of the Luxonis Data Format. A subset of `data`.     |
```
klemen1999: Link not correct, it should be to the docs.luxonis

luxonis_ml/data/README.md
```diff
@@ -1,1137 +0,0 @@
...
-```python
-for img, labels in loader:
-    ...
-```
-
-## LuxonisParser
-
-`LuxonisParser` offers a simple API for creating datasets from several common dataset formats. All of these formats are supported by `roboflow`.
-
-The supported formats are:
```
klemen1999, line 401: I think this section with expected layout is very important to be kept in the "new docs". This is the part of the docs that I most often have to link to users since it clearly explains what is the format that they need the data into be parsable by our lib.

luxonis_ml/data/README.md
```diff
@@ -1,1137 +0,0 @@
...
-
```
klemen1999, line 402: I think we should also include NATIVE layout and explanation of the annotations in here. From user PoV: I export to NATIVE format - how can I then parse this data if I don't want to use LuxonisParser, I want to parse it manually. So what are the keys, in what format are the values (e.g. what is the format of the seg mask), etc. I'm fine with linking to some other part of the docs where actual format is better explained (e.g. relevant docs in the ldf module) but at least we should have here the dir structure of NATIVE dataset.

I sppose the location for this would be under `parsers/__init__.py` in these new docs? But it should be clearly linked to from this top level `__init__.py` so I can quickly navigate to it

luxonis_ml/data/__init__.py
klemen1999, line 1: Lines 310-312 are not actually valid. It should be "ultraltyticsndjson" not "ultralytics-ndjson".
But we should rather use `native` and/or `coco` for examples here instead of using `ultralytics` 3 times.

luxonis_ml/data/parsers/__init__.py
```diff
@@ -154,39 +165,400 @@
...
+      maps each source name to a path.
+    - ``task_name`` names the task group of the annotation.
+    - ``sample_metadata`` holds record-level metadata, such as a camera name
+      or a frame number. It is not an annotation label.
+    - ``annotation`` holds one detection. The export writes one record for
+      each detection, so several records can share the same ``file``.
+
+A detection holds a ``class``, an ``instance_id``, and one payload key. The
+payload key is ``boundingbox``, ``keypoints``, ``segmentation``,
+``instance_segmentation``, ``array``, or ``metadata``. Boxes and keypoints use
```
klemen1999: Technically `array` is dropped afaik currently, we don't export those fields. It's a missing feature but that should be addressed separately.

klemen1999 (approved):
Just one more comment then I think we should be good to merge

PR #489, opened 2026-08-14
Fail the docs build on a pydoctor warning

kozlov721:
## Purpose

The docs build cannot run with `--warnings-as-errors` today. `pydoctor luxonis_ml`
reports 28 warnings on the current tree:

- 23 come from the `__main__` modules. pydoctor parses each string in an
  annotation as a forward reference. `Parameter(negative="")` and
  `ext={".json", ".yaml", ".yml"}` are not Python expressions.
- 5 come from `albumentations_engine.py`. Single backticks ask pydoctor for a
  link, and no target exists for `A.BaseCompose` and four similar names.

This PR removes all 28 warnings and makes a warning fail the build.

## Specification

Three commits, one change each:

1. `keep pydoctor from parsing CLI parameter strings` — `negative=""` becomes
   `negative=()` in the three `__main__.py` files. The `--aug-config` validator
   moves into the module constant `AUG_CONFIG_VALIDATOR`, so the file
   extensions leave the annotation.
1. `mark unresolvable Albumentations names as literals` — double backticks for
   the five external names in `albumentations_engine.py`.
1. `fail the docs build on a pydoctor warning` — `warnings-as-errors = true` in
   `[tool.pydoctor]`, plus the matching CONTRIBUTING.md text.

- No behavior change. cyclopts converts `negative=""` and `negative=()` to the
  same empty tuple.
- `ci.yaml` needs no change. `[tool.pydoctor]` holds the setting, so the local
  command and the CI job now give the same result.
- The portal export is not affected. `luxonis/docs-content` patches pydoctor to
  skip `Annotated` metadata, which is why the published CLI pages were always
  correct while the bare build warned.
- This PR stacks on #488. Merge #488 first.

## Deployment Plan

None / not applicable. The change touches docstrings, CLI parameter metadata,
and the docs configuration only.

## AI Usage

Assisted-by: Claude Code:claude-opus-5[1m]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,15 +131,21 @@ uv run pydoctor luxonis_ml
 ```
 
 Open `apidocs/index.html` to inspect the result. CI runs the same command. The
-`[tool.pydoctor]` section of `pyproject.toml` sets the Google docstring format,
-which pydoctor reads on its own.
+`[tool.pydoctor]` section of `pyproject.toml` sets the Google docstring format
+and `warnings-as-errors`, which pydoctor reads on its own.
 
 > [!IMPORTANT]
-> A malformed docstring is a syntax error. pydoctor reports it and exits
-> non-zero, so the CI job fails on broken markup. pydoctor exits 0 for
-> unresolved-reference and annotation warnings, and the build reports some of
-> those today. Read the command output after you change a docstring, because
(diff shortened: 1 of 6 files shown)


kozlov721:
@coderabbitai review

klemen1999 (approved):

PR #491, opened 2026-08-15
Named keypoints and per-task keypoint metadata in LDF records

kozlov721:
## Purpose

LDF stored keypoints as a positional list of `(x, y, visibility)` triplets. A
task's keypoint names and edges lived outside the records, in
`Metadata.skeletons`. A caller set them with `set_skeletons`. An annotation
could not say which keypoints it holds, so a partial annotation had no way to
identify them.

Horizontal flip pairs and OKS sigmas had no home at all. Nothing in LDF could
record them.

This branch lets an annotation name its keypoints and describe the keypoints of
its task. `LuxonisDataset.add` promotes the description into the dataset
metadata. The dataset thus stores one definition for each task, and no record
repeats it. That definition also holds the flip pairs and the sigmas, which are
new.

The branch also merges `main`, which added the older-LDF export target. It keeps
that target correct for keypoints.

## Specification

**Annotations**

- `KeypointAnnotation.keypoints` is a mapping from keypoint name to a triplet.
  A plain list of triplets still works. Its keypoints then take the positional
  names `"0"`, `"1"`, ....
- New `Keypoint` named tuple: `x`, `y`, `visibility`. Visibility defaults to
  `2`. It compares, unpacks and converts to `numpy` exactly like the triplet it
  replaces.
- New `KeypointMetadata` model: `labels`, `edges`, `flip_pairs`, `sigmas`. Edges
  and flip pairs accept keypoint names or indices.
- An annotation may carry `edges`, `flip_pairs` and `sigmas` as flat fields.
  These describe the task, so `add` moves them into the dataset metadata. A
  stored annotation keeps only the coordinates, in the order the task defines.

**Datasets**

- `set_keypoint_metadata` replaces `set_skeletons`. It adds `flip_pairs`,
  `sigmas` and `infer_flip_pairs`. It replaces only the fields you pass, so you
  can build a definition over several calls.
- `get_keypoint_metadata` replaces `get_skeletons`. It returns
  `dict[str, KeypointMetadata]`.
- `set_skeletons` and `get_skeletons` remain as deprecated aliases. They forward
  to the new names and emit a `DeprecationWarning`.
- `Metadata.skeletons` becomes `Metadata.keypoint_metadata`. The old key stays a
  validation alias, so a dataset written earlier still opens.
- When you give no flip pairs, `add` infers them from `left` and `right` names.
  The match is deliberately narrow, because a wrong pair mirrors the wrong
  keypoints and never fails.
- A merge warns when two datasets describe different keypoints for one task.
- `LDF_VERSION` moves from 2.1 to 2.2.

**Native export (merge with `main`)**

- `NativeExporter` accepts `keypoint_metadata` and `ldf_version`. The export
  writes the task fields once for each task and split. `NativeParser` gives them
  back to `LuxonisDataset.add`.
- `LDFDowngrader` now drops annotation fields as well as record fields. Without
  this, an export stamped LDF 2.0 still carried the 2.2 keypoint fields. An
  annotation forbids extra fields, so the older install rejected the record.
- New export target LDF 2.1. It keeps `sample_metadata`. It drops the keypoint
  names, the edges, the flip pairs and the sigmas. Supported targets are now
  2.2, 2.1 and 2.0.

No new dependencies.

Breaking changes:

- `KeypointAnnotation.keypoints` is a mapping, not a list. `to_numpy()` and
  `combine_to_numpy()` give the same arrays as before.
- `get_skeletons()` returns `dict[str, KeypointMetadata]`. It returned
  `dict[str, tuple[list[str], list[tuple[int, int]]]]`. The call still works,
  but it now warns. Use `get_keypoint_metadata()`.
- `luxonis_ml.data.datasets.metadata.Skeletons` is gone. `luxonis_ml.data` never
  exported it, so only a deep import breaks. Use
  `luxonis_ml.ldf.KeypointMetadata`.
- `LDF_VERSION` 2.2 means an export made now needs `ldf_version="2.1"` or
  `"2.0"` to stay readable by an older luxonis-ml.

`luxonis-train`, `luxonis-eval` and `modelconverter` consume `luxonis_ml.data`.
They pin published versions, so they see this only at the next release. One
known break waits there. `luxonis_train/loaders/luxonis_loader_torch.py:226`
does `len(skeletons[task][0])`, and a `KeypointMetadata` model has no index. The
fix is `self.dataset.get_n_keypoints()`, which already gives that number.
`luxonis-eval`, `modelconverter` and `datadreamer` do not use skeletons.

A dataset written before this change still loads. The stored definition holds
`labels` and `edges` only, and the new model treats the other fields as empty.
`test_a_legacy_dataset_still_loads` covers this.

## Deployment Plan

None / not applicable. This is a library change with no service to roll out. It
ships in the next `luxonis-ml` release, and a revert is a plain revert of the
branch.

## Testing & Validation

New tests: `tests/test_ldf/test_keypoints.py`,
`tests/test_data/test_keypoint_metadata.py`, and the export-version tests in
`tests/test_data/test_export_ldf_version.py`.

The downgrade fix and the LDF 2.1 target each ship regression tests. Both were
confirmed to fail without their fix.

Run locally on this branch:

- 188 tests pass: `tests/test_ldf`, `test_keypoint_metadata.py`,
  `test_annotations.py`, `test_export_ldf_version.py`,
  `test_dataset_metadata.py` and `tests/test_data/test_utils`.
- 38 doctests pass across `luxonis_ml/ldf` and `luxonis_ml/data`.
- `pyright --warnings --level warning --project pyproject.toml` reports 0 errors
  and 0 warnings.
- `pre-commit` is clean on every touched file.

Not run locally: the tests whose fixtures download from
`gs://luxonis-test-bucket`. The local Google credential is not valid, and each
of those tests fails on the download. `tests/test_data/test_dataset.py` gives
the same result on this branch as on the commit before it, so the branch adds no
failure. CI must cover them.

## AI Usage

Assisted-by: Claude Code:claude-opus-5

Seven of the nine commits carry a `Co-Authored-By` trailer. The assistance
covers:

- the merge with `main`;
- the LDF downgrade fix and the LDF 2.1 export target;
- the rename from keypoint skeletons to keypoint metadata;
- a de-AI review of the whole branch;
- the pass that made every keypoint metadata name explicit.

Commits `67b1285` and `9c0280d` are hand written.

## Summary by CodeRabbit

- **New Features**
  - Added structured keypoint metadata for labels, edges, flip pairs, sigmas, and inference settings.
  - Added keypoint metadata access, validation, alignment, padding, and improved annotation loading.
  - Added metadata support to native and COCO exports.
  - Added LDF 2.1 export compatibility and support for newer keypoint fields.

- **Bug Fixes**
  - Improved dataset merging across compatible LDF versions while rejecting incompatible major versions.
  - Improved keypoint visualization and inspection.

- **Documentation**
  - Expanded data, parser, annotation, export, and keypoint metadata documentation.


luxonis_ml/data/datasets/metadata.py
```diff
@@ -151,3 +157,53 @@ def _sort_classes(self, classes: Iterable[str]) -> list[str]:
...
+
+    """
+    if mine.labels != theirs.labels:
+        logger.warning(
+            f"Task '{task}' has different keypoint labels in the two "
+            "datasets being merged. Keeping the ones from the dataset "
+            "being merged in. They now describe the keypoints of the "
+            "other dataset too, which gave them a different order."
+        )
+        return theirs
```
klemen1999: [P1] Please reindex the stored keypoint rows or reject this merge when label orders differ. `LuxonisDataset.merge_with` concatenates the positional Parquet rows unchanged, but this branch assigns the incoming labels to both datasets. I reproduced a merge of `["nose", "left_eye"]` with `["left_eye", "nose"]`: the original nose coordinate is then interpreted as `left_eye`. The warning does not prevent the resulting annotation corruption.

Note: This is an AI review.

luxonis_ml/ldf/annotation.py
```diff
@@ -852,31 +1359,93 @@ def combine_to_numpy(
...
+        annotation = self
+        if keypoint_metadata is not None:
+            aligned = keypoint_metadata.align(self.keypoints)
+            # Compared as items, not as mappings. Dict equality ignores
+            # order, and the stored payload encodes only the order.
+            if list(aligned.items()) != list(self.keypoints.items()):
+                annotation = self.model_copy(update={"keypoints": aligned})
+        return annotation.model_dump_json()
+
+    @model_serializer(mode="plain", when_used="json")
```
klemen1999: [P2] Could the positional-only representation be limited to `to_parquet_json`? This model serializer also runs for ordinary `KeypointAnnotation.model_dump_json()` and nested `DatasetRecord.model_dump_json()`. I reproduced a JSON round trip of a record with named keypoints, edges, and sigmas; after validation, `declared_metadata()` is `None` because all of those fields were removed. Standard JSON serialization should retain the full annotation definition.

Note: This is an AI review.

PR #493, opened 2026-08-16
Fix the background class of a segmentation task

kozlov721:
## Purpose

Two defects in how `LuxonisLoader` handles the synthesized `background` class.

**1. A sample with no mask claimed no class at all.** When a task's masks leave
pixels unassigned, the loader synthesizes a `background` class and assigns every
unclaimed pixel to it. That fill reached only a sample that carried a mask:

| Sample | background | dog | Pixels with a class |
| --- | --- | --- | --- |
| has a mask over the top half | 8 | 8 | 16 of 16 |
| has no mask at all | 0 | 0 | **0 of 16** |

The second row is not a valid per-pixel target. A softmax over the class
channels has no positive channel to point at, so the sample teaches the model
nothing, where the first row teaches it "the bottom half is background".

**2. The class map depended on the split order.** The loader decided whether a
task needs a background class by reading **sample 0 alone**. The same dataset
therefore returned a different number of channels depending only on which sample
the shuffle put first:

| Unannotated sample first | Classes | Mask shape |
| --- | --- | --- |
| yes | `cat, dog` | `(2, 32, 32)` |
| no | `background, cat, dog` | `(3, 32, 32)` |

## Specification

An empty segmentation label is filled with the background class, exactly as an
annotated one is.

`LuxonisLoader._uncovered_tasks` replaces the single-sample probe. It reads
every sample of the view. For each sample it merges the masks of a task into
their union, with `pycocotools.mask.merge`, then compares the area of that
union with the area of the image. `pycocotools` works on the run-length
encoding itself, so the check decodes no mask and opens no image. One uncovered
sample settles a task, so the scan stops at the first one.

The union counts a pixel once, however many masks claim it.
`SegmentationAnnotation.combine_to_numpy` also assigns a pixel once, and gives
it to the first mask that claims it. The check and the loader output therefore
agree on the pixels that a task leaves free.

LDF 1.0 stored a polyline in place of an encoded mask, and no mask size beside
it. The points of a polyline are normalized, so `_sample_masks` rasterizes them
on a grid of its own. It needs no image to learn the size.

### Alternatives considered

**Always add a background class** to any segmentation task that does not define
one. No scan at all, no approximation, and no order to depend on. It was not
taken because it changes the output shape of every fully covered dataset, which
gains a channel that is always empty, and a model then spends capacity on a
class that never appears. The scan keeps the current meaning of the class and
only makes the decision reproducible.

**Record the answer at ingestion.** `LuxonisDataset.add` already touches every
mask, so it could store per-task coverage in `metadata.json` and leave the
loader nothing to compute. This is the better long-term answer, and it is not in
this PR because it needs an LDF metadata field, a way to update it as `add` is
called again, and a fallback for every dataset written before it existed.

**Make it explicit.** A loader or dataset setting, rather than any inference at
all. Worth having beside one of the above, but it does not answer what to do
with the datasets that exist today.

Both changes affect what a model is trained on, in a dataset where the loader
synthesizes a background class:

- An unannotated sample now contributes a full background target instead of an
  empty one.
- A dataset whose first sample happened to cover every pixel, or to carry no
  mask, now gains the background class it always needed. Its segmentation output
  gains a channel.

Datasets that name their own `background` class at index 0 are unaffected, as
are tasks whose masks already cover every pixel.

### Performance

The loader decides once, when it opens. `main` decodes image 0 to decide, so it
pays for the resolution. This PR reads the encodings, so it pays for the rows.

| Samples | Image | Masks of a sample | `main` | this PR |
| ---: | :--- | :--- | ---: | ---: |
| 200 | 256×256 | leave pixels free | 1.4 ms | ⚪ 1.5 ms |
| 1000 | 256×256 | leave pixels free | 1.4 ms | 🔴 3.4 ms |
| 200 | 1024×1024 | leave pixels free | 24 ms | 🟢 2.3 ms |
| 200 | 256×256 | cover every pixel | 1.6 ms | 🔴 7.3 ms |
| 1000 | 256×256 | cover every pixel | 1.4 ms | 🔴 29 ms |
| 3000 | 256×256 | cover every pixel | 1.4 ms | 🔴 73 ms |
| 200 | 1024×1024 | cover every pixel | 28 ms | 🟢 16 ms |

🟢 faster than `main` · 🔴 slower · ⚪ inside the noise. Median of 21
measurements, with a new loader for each. 3 masks per sample.

Large images win, because this PR opens no image. Small images that cover every
pixel lose, and the loss grows with the dataset: 73 ms once, at 3000 samples.

The union costs about 15% more than a sum of the mask areas. A sum is wrong:
two masks that overlap cover fewer pixels than their areas add up to.

<details>
<summary>The script that produced those numbers</summary>

```python
"""Cost of the decision whether a segmentation task needs a background class.

The loader makes the decision once, so every measurement gets a new loader. A
best-of-many measurement on one loader would report a warm polars query that
no loader ever gets.

Run the script on this branch and on `main`, then compare.

    python bench_background_probe.py /tmp/probe-bench
"""

import os
import statistics
import sys
import time
from pathlib import Path

import cv2
import numpy as np

BASE = Path(sys.argv[1] if len(sys.argv) > 1 else "/tmp/probe-bench")
BASE.mkdir(parents=True, exist_ok=True)
os.environ["LUXONISML_BASE_PATH"] = str(BASE / "luxonisml")

from luxonis_ml.data import LuxonisDataset, LuxonisLoader  # noqa: E402

CLASSES = ["road", "sky", "tree"]
# Samples, image size, and whether the masks cover every pixel.
CASES = [
    (200, 256, False),
    (1000, 256, False),
    (200, 1024, False),
    (200, 256, True),
    (1000, 256, True),
    (3000, 256, True),
    (200, 1024, True),
]
ON_THIS_BRANCH = hasattr(LuxonisLoader, "_uncovered_tasks")


def image(size: int, index: int) -> Path:
    folder = BASE / f"images_{size}"
    folder.mkdir(exist_ok=True)
    path = folder / f"img_{index}.jpg"
    if not path.exists():
        rng = np.random.default_rng(index)
        cv2.imwrite(
            str(path), rng.integers(0, 255, (size, size, 3), dtype=np.uint8)
        )
    return path


def partition(size: int, seed: int, covered: bool) -> np.ndarray:
    """Label every pixel 0, 1 or 2, under two wavy horizons.

    An uncovered partition shrinks each region, which opens a gap between
    them. Two straight horizons would give the masks a much shorter
    run-length encoding than a real dataset has.
    """
    rng = np.random.default_rng(seed)
    x = np.arange(size)

    def horizon(centre: float) -> np.ndarray:
        curve = sum(
            rng.normal() * np.sin(2 * np.pi * k * x / size + rng.uniform(0, 7))
            for k in range(1, 6)
        )
        return np.clip(centre + curve * size / 40, 1, size - 2)

    first = horizon(size / 3)
    second = np.maximum(horizon(2 * size / 3), first + 2)
    rows = np.arange(size)[:, None]
    label = np.zeros((size, size), dtype=np.uint8)
    label[rows >= first[None, :]] = 1
    label[rows >= second[None, :]] = 2
    if not covered:
        gap = size // 16
        label[(rows >= first[None, :]) & (rows < first[None, :] + gap)] = 3
        label[(rows >= second[None, :]) & (rows < second[None, :] + gap)] = 3
    return label


def build(samples: int, size: int, covered: bool) -> LuxonisDataset:
    name = f"probe_{samples}_{size}_{'covered' if covered else 'uncovered'}"
    if LuxonisDataset.exists(name):
        return LuxonisDataset(name)

    def generator():
        for i in range(samples):
            path = image(size, i)
            label = partition(size, i, covered)
            for m in range(3):
                yield {
                    "file": str(path),
                    "task_name": "scene",
                    "annotation": {
                        "class": CLASSES[m],
                        "segmentation": {
                            "mask": (label == m).astype(np.uint8)
                        },
                    },
                }

    LuxonisDataset(name).add(generator()).make_splits((1.0, 0.0, 0.0))
    return LuxonisDataset(name)


def _once(call) -> float:
    start = time.perf_counter()
    call()
    return time.perf_counter() - start


def decision(loader: LuxonisLoader, classes: dict):
    """Return the branch's own background-class decision, alone."""
    if ON_THIS_BRANCH:
        return loader._uncovered_tasks

    from luxonis_ml.data.utils import get_task_name, task_type_iterator

    def probe_sample_zero() -> None:
        _, labels, _ = loader._load_data(0)
        for task, masks in task_type_iterator(labels, "segmentation"):
            names = classes[get_task_name(task)]
            if masks.shape[0] > 1 and names.get("background") != 0:
                np.any(np.sum(masks, axis=0) == 0)

    return probe_sample_zero


print(f"{'samples':>8s} {'image':>10s} {'masks':>10s} {'decision':>11s}")
for samples, size, covered in CASES:
    dataset = build(samples, size, covered)
    # The class map as it stands before the loader adds `background`.
    classes = dataset.get_classes()
    decide = statistics.median(
        _once(decision(LuxonisLoader(dataset, view="train"), classes))
        for _ in range(21)
    )
    print(
        f"{samples:8d} {f'{size}x{size}':>10s} "
        f"{'covered' if covered else 'uncovered':>10s} "
        f"{decide * 1000:8.2f} ms",
        flush=True,
    )
```

</details>

## Deployment Plan

None / not applicable

## Testing & Validation

`tests/test_data/test_background_class.py` holds one test per defect, and both
fail on `main`:

- `test_a_sample_with_no_mask_is_all_background` asserts that every pixel of
  every sample belongs to exactly one class.
- `test_the_background_class_ignores_the_split_order` loads one dataset twice,
  with the unannotated sample first and last, and asserts the class maps match.
  On `main` it fails with `{'cat': 0, 'dog': 1} == {'background': 0, 'cat': 1,
  'dog': 2}`.

Three more tests guard the new check:

- `test_overlapping_masks_do_not_hide_uncovered_pixels` gives one image two
  masks of 300 rows each, of the 512 rows an image holds. Their areas add up to
  307200, above the 262144 the image holds, while their union stops at row 400.
  A check that adds the areas calls the task covered, and drops the background
  class. `main` passes this test, because it decodes the masks and so counts
  each pixel once. The union keeps that property, and decodes nothing.
- `test_a_polyline_mask_needs_no_image_to_measure` and
  `test_a_polyline_takes_the_size_of_the_mask_beside_it` call `_sample_masks`
  with the LDF 1.0 payload, which carries no mask size.

`test_empty` asserted the all-zero mask. It now asserts that no foreground class
is set, which is the invariant it was written for. It runs through `Mosaic4`,
which pads the parts no source image reaches, so it cannot assert an exact pixel
total.

`tests/test_data/test_augmentations` and `test_background_class.py` give 360
passed. The loader, annotation, class-ordering, task-ingestion, multi-input and
LDF suites give 68 passed and 4 failed. The same 4 fail without this change, and
every one is a Google credential error in this environment.
`pyright --warnings --level warning --project pyproject.toml` reports no errors,
and `pre-commit run --all-files` passes.

## AI Usage

Assisted-by: Claude:claude-opus-5

Submitted code was reviewed by a human: NO

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved segmentation analysis to reliably detect uncovered image areas.
  - Added support for encoded masks and normalized polygon annotations without loading images.
  - Automatically fills empty segmentation regions with the appropriate background class.
  - Preserves uncovered background pixels when masks overlap and maintains consistent class mappings.
- **Tests**
  - Expanded coverage for empty samples, background classes, mask overlaps, split ordering, and mixed annotation formats.


kozlov721:
Both review findings are addressed.

**Split-order-dependent background detection.** Confirmed and fixed. The same dataset returned `(2, 32, 32)` or `(3, 32, 32)` masks depending only on which sample the shuffle put first. `_uncovered_tasks` now reads every sample of the view, counting set pixels from the run-length encoding rather than decoding the mask, and stops at the first uncovered sample. The test no longer reorders the split to work around the behaviour; it reorders it to prove the result no longer depends on the order.

**Assert coverage for each pixel.** Taken, with a stronger assertion than suggested. `combine_to_numpy` masks out the pixels a class already claimed, so no pixel can carry two classes and the guarantee is *exactly* one, not at least one. The test asserts `np.array_equal(mask.sum(axis=0), np.ones((512, 512)))`.

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -383,6 +367,54 @@ def __getitem__(self, idx: int) -> LoaderOutput:
...
+    def _uncovered_tasks(self) -> set[str]:
+        """Return the segmentation tasks that leave pixels unassigned.
+
+        The check reads every sample of the view. A check on one sample
+        lets the split order decide the class count of the task.
+
+        `pycocotools.mask.area` counts the set pixels of the run-length
+        encoding, which costs much less than a decode. A pixel under two
+        masks counts twice, so a task with much overlap can pass the
+        check. The old check on sample 0 made the same approximation.
```
klemen1999: don't mention the "old way"

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -383,6 +367,54 @@ def __getitem__(self, idx: int) -> LoaderOutput:
...
+                continue
+            covered = 0
+            size = 0
+            for annotation in annotations:
+                data = json.loads(annotation)
+                # LDF 1.0 stored polylines, which carry no size to compare.
+                if "counts" not in data:  # pragma: no cover
+                    continue
+                size = data["height"] * data["width"]
+                covered += int(
```
klemen1999: We are summing up individual masks but the problem could be overlaps. If two masks overlap then the size check can pass but pixels are still uncovered. We should do this check on the union of all masks per image not on individual annotations. Not sure how this affects the performance though

luxonis_ml/data/loaders/luxonis_loader.py
```diff
@@ -383,6 +367,54 @@ def __getitem__(self, idx: int) -> LoaderOutput:
...
+        for _, task_name, annotations in samples.iter_rows():
+            # One uncovered sample settles the task.
+            if task_name in uncovered:
+                continue
+            covered = 0
+            size = 0
+            for annotation in annotations:
+                data = json.loads(annotation)
+                # LDF 1.0 stored polylines, which carry no size to compare.
+                if "counts" not in data:  # pragma: no cover
```
klemen1999: So now LDF 1.0 won't work right? Can we do something about this? In line 547 we do still handle the 1.0 LDF so maybe we can do something similar here?

klemen1999 (approved):

